PR #4, opened 2015-01-31
Return empty string; Fix null crash; Add tests; camelCase

### LINKED ISSUE #3: Some issues to make this compatible with React addon classSet
1. `classnames('a', false)` produces `'a false'` in classSet, but classnames ignores `false`. Passing `null` crashes it. Passing `undefined` gets it ignored. Doesn't matter what the desirable behaviour is; it breaks compatibility with classSet.

2. Bikeshedding: I know this has nothing to do with React, but React uses `className` so it'd be nice if this could be `classNames` camelCased.

3. Empty class names return `undefined` rather than `''`.

chenglou:
Fixes #3 (didn't feel like submitting individual PRs)

- Return empty string rather than `undefined`. Better behaviour imo and good for type checking in the future.
- camelCase `classNames`. This doesn't affect anything since the user can rename it anyways during `require`. Just thought it'd be more consistent.
- Make the function not crash on `null`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,15 +1,20 @@
-function classnames() {
+function classNames() {
 	var args = arguments, classes = [];
 	for (var i = 0; i < args.length; i++) {
-		if (args[i] && 'string' === typeof args[i]) {
-			classes.push(args[i]);
-		} else if ('object' === typeof args[i]) {
-			classes = classes.concat(Object.keys(args[i]).filter(function(cls) {
-				return args[i][cls];
+		var arg = args[i];
+		if (arg == null) {
+			continue;
+		}
+
+		if ('string' === typeof arg) {
+			classes.push(arg);
+		} else if ('object' === typeof arg) {
+			classes = classes.concat(Object.keys(arg).filter(function(cls) {
+				return arg[cls];
 			}));
 		}
 	}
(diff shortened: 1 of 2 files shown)


chenglou:
Man I wish I could do stacked PRs a-la Phabricator. @JedWatson after this I'll refactor the code to make it as allocation-free as possible.

index.js
```diff
@@ -1,15 +1,20 @@
...
+		if ('string' === typeof arg) {
+			classes.push(arg);
+		} else if ('object' === typeof arg) {
+			classes = classes.concat(Object.keys(arg).filter(function(cls) {
+				return arg[cls];
 			}));
 		}
 	}
-	return classes.join(' ') || undefined;
+	return classes.join(' ') || '';
```
akre54: `[].join(' ') => ''`. No need for the guard here

JedWatson:
Thanks. I merged and fixed the guard so you can get on with the next refactor. Would be good to publish a new version with the changed behaviour, what's your opinion on new version number to comply with semver?

I'm thinking 1.1.0 - these don't seem to really be _breaking_ changes...

chenglou:
Yeah sounds fine. Lemme get the perf refactor in though (slightly busy here)

PR #6, opened 2015-02-06
Make code uglier and faster; more precise test

### LINKED ISSUE #5: Makes this much faster
Sorry, just filing an issue for myself. Really busy these days but will get to it.

chenglou:
The function ignores non-string values. This has always been the case. No behavior change here.

Addtionally, the function's now consistently 4x faster due to less allocations. I kept the allocation for `classNames`, because it turns out that it's faster to mutate an array and join it at the end than to continuously append to a string, especially for slightly longer class names.

Fixes #5

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,7 @@
 function classNames() {
-	var args = arguments, classes = [];
+	var args = arguments;
+	var classes = [];
+
 	for (var i = 0; i < args.length; i++) {
 		var arg = args[i];
 		if (arg == null) {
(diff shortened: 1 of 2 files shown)


chenglou:
@JedWatson This is probably it. I can move the readme from https://github.com/petehunt/react-classset here and deprecate that repo completely (there are slight differences regarding non-string values but nobody cares).

JedWatson:
:+1: thanks @chenglou

PR #8, opened 2015-02-13
Fix behavior for non-zero params & empty string

chenglou:
A bit trivial but this was in discordance with the README, which stated
that "falsy keys won't be included". Previously `1` was ignored.

Same for empty string. `c('a', '')` gave `"a "`. Fixed that.

P.S. I hate JS implicit conversion and falsy values


JedWatson:
Good catch, @chenglou

PR #10, opened 2015-02-25
Simplify loop to only the asserted case

ssorallen:
This prevents needing to use `continue` and explicitly defines the case
where the class should be added.


JedWatson:
Thanks @ssorallen, makes sense to me.

PR #11, opened 2015-02-25
package: remove unnecessary pathing

dcousens:
npm automatically includes `~/node_modules/.bin` in the PATH when running scripts; so there is no need to specify this.

It will not revert to the globally installed `mocha` unless the package is missing.


JedWatson:
Thanks for the simplification @dcousens

Issue #12, opened 2015-02-25
Take array as argument?


davidtheclark:
Readme clearly explains how to use an array ("if you have an array of these, use apply"). But I wonder if there's any reason not to also accept an array as a regular argument. Would you be open to a PR enabling that?

JedWatson:
What would the array do?

I've been considering an array-based syntax to allow dynamic class names to be conditionally added. It's  bit of a weird syntax but one that would solve some otherwise awkward code.

e.g.

```
var a = 'variable-class-name';
var b = true;
classNames('static-class-name', [a, b]) === 'static-class-name variable-class-name';
```

That would, of course, be incompatible with using arrays as lists of classNames to include, which is what I think you're suggesting:

```
classNames(['a', 'b'], 'c') === 'a b c';
```

... however I think it would be more useful. Feedback welcome.

davidtheclark:
Yes, what I was suggesting is just using arrays as values. For one reason or another you might build up a list of classnames as an array and this function could then be used to apply them. Here's are two simple examples:

```js
// conditionally build array of dynamic class names
var arr = ['a', 'b'];
if (c) arr.push('super' + c);
classNames(arr);

// use an array of classnames that is stored as data somewhere
var prefabArr = require('./classesForThing');
classNames(prefabArr.push('newone'));
```

Both of these situations (and others like them) *could* be addressed by dynamically building an object with `true` values. (The array is just more succinct.)

```js
var obj = { a: true, b: true };
if (c) obj['super' + c] = true;
classNames(obj);
```

The same could be said for the array syntax you were considering. In fact, it seems especially appropriate in that case because you are essentially doing the same thing that an object does. Why not this instead (?):

```js
classNames({
  'static-class-name': true,
  'variable-class-name': b
});
```

tgriesser:
Just came here to open the same ticket, I'd be +1 on using arrays as values (and recursively, so you don't need to worry about flattening them).

In ES6 (via Babel) computed properties are valid so the `[key, bool]` array syntax eventually won't be necessary:

```js
classNames({
  'static-class-name': true,
  [`variable-class-${name}`]: b
});
```

JedWatson:
> In ES6 (via Babel) computed properties are valid so the [key, bool] array syntax eventually won't be necessary

Good point. PR is in review.

chenglou:
Humm I'm somewhat not in favour of this, for a few reasons.

[...]

PR #13, opened 2015-03-08
Accept array arguments

davidtheclark:
In case you are interested in accepting array arguments, as requested in https://github.com/JedWatson/classnames/issues/12, here's a PR that adds the functionality along with tests and a little documentation.

Instead of just inserting another `if` statement to `index.js` I made some significant revisions. If you don't like them, I can back out, of course. But my reasoning was that to get the intended functionality (recursive array checking) I needed (1) a named function that could be called recursively, and (2) a reliable way to test if an argument is an array. This PR uses an array-testing method I've seen some reliable people rely on, the one used by [lodash](https://github.com/lodash/lodash/blob/3.4.0/lodash.src.js#L8684) and [Underscore](http://underscorejs.org/docs/underscore.html#section-137); and since the method is equally useful to check the other types that needed checking, I thought I'd work in a simple switch statement.

Please let me know what you think, if you want anything changed, etc. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ Install with npm, or download the [UMD version](http://wzrd.in/standalone/classn
 npm install classnames
 ```
 
-The `classNames` function takes any number of arguments which can be a string or object.
+The `classNames` function takes any number of arguments which can be strings, objects, or arrays that contain strings, objects, and other arrays.
 The argument `'foo'` is short for `{foo: true}`. If the value of the key is falsy, it won't be included in the output.
 
 ```js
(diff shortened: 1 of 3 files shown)


k-ori:
+1

JedWatson:
@chenglou you made some specific changes to ensure the performance of this package, do you have any feedback on the implications of the changes @davidtheclark made?

ZeikJT:
Why is the `checkArg` function inside of the `classNames` function? Doing that forces a new function to be created each time someone runs `classNames`, it's best if you move it out and add `classes` to the params. It doesn't need anything else from scope so it should be an easy change.

Also, `hasOwnProperty` is relatively expensive, it might be more performant to use `Object.keys`

JedWatson:
This functionality was added along with a major speed improvement in #16

PR #16, opened 2015-03-17
Adds support for Arrays and Performance Improvements

danielstreit:
Here's another implementation to support Arrays. This also removes the use of Array#join in favor of String concatenation. [JSPerf] (http://jsperf.com/classnames-util/3)


JedWatson:
Killer. That's a huge improvement. Thanks!

PR #24, opened 2015-04-04
Use Array.isArray to check for array

stuartsan:
Array.isArray is faster than checking using Object.prototype.toString and is supported in modern browsers. React requires es5-shim.js (which includes Array.isArray) to support IE8. I modified the tests from #16 for benchmarks; pretty decent improvements shown.

http://jsperf.com/classnames-util/5

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,11 @@ var arr = ['b', { c: true, d: false }];
 classNames('a', arr); // => 'a b c'
 ```
 
+## Polyfills needed to support older browsers
+
+`Array.isArray`: see [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray)
+ for details about unsupported older browsers (e.g. <= IE8) and a simple polyfill.
+
 ## License
 
 (The MIT License)
--- a/index.js
+++ b/index.js
@@ -10,7 +10,7 @@ function classNames() {
 
 		if ('string' === typeof arg || 'number' === typeof arg) {
 			classes += ' ' + arg;
-		} else if (Object.prototype.toString.call(arg) === '[object Array]') {
+		} else if (Array.isArray(arg)) {
 			classes += ' ' + classNames.apply(null, arg);
 		} else if ('object' === typeof arg) {
 			for (var key in arg) {


dcousens:
+1 from me.  Will let @JedWatson merge

hartzis:
+1 - Thanks for the addition and the jsperf to prove it! Already merged into my local usage.

longlho:
if we drop IE8 support I'm down, but it's a pretty big IF.

dcousens:
From https://github.com/JedWatson/classnames/issues/25

@longlho is the alternative suitable to avoid that pitfall?

`object != null && object.constructor === Array`

longlho:
@dcousens I'm down w/ that :+1:

hartzis:
Updated the jsperf tests to include both of the changes being talked about, also added tests that include arrays, which seems relevant.

http://jsperf.com/classnames-util/6

Looks like `object.constructor` pulls ahead, but do you actually need the `object != null`?

dcousens:
Only if you can guarantee the argument isn't null,  in case,  no you don't.

dcousens:
In this case*

stuartsan:
If the project must support IE8 without any shims, I think the existing implementation, using Object.prototype.toString, is already the most robust way to check.

Using instanceof Array or thing.constructor === Array can fail in the case of documents with multiple frames, because each frame's constructors have their own identities.

dcousens:
@stuartsan is that a common concern?  I feel like so many other assumptions would probably break in such cases.

dcousens:
I'm still +1 for this.  @JedWatson ?

stuartsan:
@dcousens yeah, I'm not sure honestly. It's common enough that some libraries still use it (e.g. https://github.com/lodash/lodash/blob/master/lodash.js#L8455 ) and many used to. Then again jQuery these days just uses Array.isArray (https://github.com/jquery/jquery/blob/154166458284bcce7d6a86328b7fd13483232a1a/src/core.js#L200) ... I'd still personally vote for using that, if unshimmed IE8 doesn't have to be supported (as in React).

dcousens:
It's literally the most canonical way to do it, provided to us by the
browser's.  I see no reason why it shouldn't be that we use,  at the least.
I'm happy to merge this.

JedWatson:
I'm happy to merge it if we add docs to the Readme explaining the requirement for an `Array.isArray` shim in IE8 (and link to / provide one that can be used)

@stuartsan do you mind adding that to your PR?

stuartsan:
@JedWatson yep no problem, added a link to MDN's page on the method, which provides a simple polyfill.

I considered putting a note to the effect of, "if you're supporting React in older browsers you're already covered" since React requires shims for several methods including this, but I'm not sure if y'all want to refer specifically to React so I left it out.

JedWatson:
Thanks @stuartsan, that's great. Will push a new version with this now.

PR #26, opened 2015-04-14
Minor addition to README to make explicit how spaces are handled in clas...

cymen:
...s string

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ The argument `'foo'` is short for `{foo: true}`. If the value of the key is fals
 
 ```js
 classNames('foo', 'bar'); // => 'foo bar'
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,6 +16,7 @@ The argument `'foo'` is short for `{foo: true}`. If the value of the key is fals
 
 ```js
 classNames('foo', 'bar'); // => 'foo bar'
+classNames('foo bar', 'bar'); // => 'foo bar'
```
dcousens: We don't have de-duping, shouldn't that be `foo bar bar`?

cymen: Whoops -- I was looking at two modules and I got them reversed. That was exactly what I was trying to document but I was wrong. Thanks!

cymen:
Closing due to misunderstanding.

PR #28, opened 2015-04-20
Add contents of LICENSE to top of index.js

wwcline:
Want the license header to be preserved when this library is
included in a project.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,3 +1,26 @@
+/*
+The MIT License (MIT)
+
+Copyright (c) 2015 Jed Watson
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+*/
 function classNames() {
 	var classes = '';
 	var arg;


dcousens:
@wwcline is this really necessary?

The LICENSE is part of the project and is found in the projects source directory (top level).
What makes this file level distinction any better than the project level distinction that currently exists?

JedWatson:
I'm guessing this is useful when the `index.js` file is included in a project without the rest of the files, in order to comply with the terms of the license (to keep copyright intact)..?

I'll modify it to include the copyright, the MIT License note and URL of the full LICENSE on github so it's not such a large header block. That should suffice.

dcousens:
It is,  but at that point is at the responsibility of the user to attach
the license.

dcousens:
To quote:

> The above copyright notice and this permission notice shall be included
in all copies or substantial portions of the Software

This burden is on the user who is using a substantial portion.  Not the
original author.

JedWatson:
@dcousens I simplified it to

```
/*
  Copyright (c) 2015 Jed Watson.
  
  Licensed under the MIT License (MIT), see
  https://github.com/JedWatson/classnames/blob/master/LICENSE
*/
```

seems a reasonable enough change if it simplifies the inclusion of the code in a project... I'm guessing we're catering for custom workflows that don't use npm or bower here.

nkbt:
When you include a library you will minify and strip all comments anyway. That does not make too much sense. Properly added license - `package.json` + `LICENSE` [+ `README`] should be sufficient.
Then you simply need a script to extract all licenses from your dependencies and whip them with your product.

Example npm packages to extract licenses:
- license-checker
- license-report
- license-spelunker 
- licensing
- npm-license 

We ended up with `license-spelunker` + little custom script to format results properly.

License headers are pretty hard to distinct and be extracted automatically.

dcousens:
@JedWatson that's fine, but it is unnecessary; and it typically indicates that that file is under a different license to what the project is licensed under.

JedWatson:
Thanks for the feedback guys.

@dcousens I got that trick from React, which includes this in every file:

```
/**
 * Copyright (c) 2015-present, Facebook, Inc.
 * All rights reserved.
 *
 * This source code is licensed under the BSD-style license found in the
 * LICENSE file in the root directory of this source tree. An additional grant
 * of patent rights can be found in the PATENTS file in the same directory.
 */
```

Their version is arguably clearer in that it specifies the LICENSE file in the project itself...

I like @nkbt's suggestion of using npm packages to extract licenses but am not clear for which environment this was an issue in the first place.

@wwcline can you shed some more light?

nkbt:
We use this way in Arbor Networks and it seems like working ok from lawyer's
point of view.



-- 
Sincerely yours, Nikita Butenko
Google Talk: nik@butenko.me
Skype: nikita.butenko
Mobile: +61 450 264 801

wwcline:
> I'm guessing this is useful when the index.js file is included in a project without the rest of the files, in order to comply with the terms of the license (to keep copyright intact)..?

@JedWatson, this is exactly what I'm trying to do, yes. My build is configured to preserve license headers, so it will keep the license header intact while minifying the rest. The simplified version of the license that you suggested  would be perfect:

```
/*
  Copyright (c) 2015 Jed Watson.

  Licensed under the MIT License (MIT), see
  https://github.com/JedWatson/classnames/blob/master/LICENSE
*/
```

I sympathize with @dcousens not wanting to duplicate what's already in the LICENSE file. My Rails app's asset pipeline uses Browserify (via the browserify-rails RubyGem) to include npm packages into its JavaScript, and I haven't found a way to configure it to prepend the LICENSE file when it processes a package.

I see you merged this request — thank you! — but I will check out @nkbt's suggested license-extraction packages, which might help me out the next time I encounter this with another library.

JedWatson:
@wwcline no problem :smile:

Daniel15:
> When you include a library you will minify and strip all comments anyway.

Putting an exclamation mark or `@preserve` at the start of the docblock will preserve the comment with most minifiers:
```
/*!
  Copyright (c) 2015 Jed Watson.

  Licensed under the MIT License (MIT), see
  https://github.com/JedWatson/classnames/blob/master/LICENSE
*/
```

Generally you'd do that with one of the docblock comments in a library, so there's a single license comment in the minified output.

nkbt:
@Daniel15 I know that this is technically doable. But imagine 200 bytes lib size and additional couple of kilobytes of license text (which I've seen not once) - that is a little bit unfair for the end-user. All non-working code must be stripped away from the production build.

Daniel15:
>  But imagine 200 bytes lib size and additional couple of kilobytes of license text (which I've seen not once)

Just a short version that says "Licensed under the MIT License" is fine, and gzipping makes such small increases relatively unimportant anyways. It makes the license more obvious to anyone looking at the minified version, which I think is necessary for code released publicly.

Pretty much every major library does this in their minified versions, see [React](http://cdn.jsdelivr.net/react/0.13.2/react.min.js) and [jQuery](http://cdn.jsdelivr.net/jquery/2.1.3/jquery.min.js) for example. jQuery's one is simply:
```
/*! jQuery v2.1.3 | (c) 2005, 2014 jQuery Foundation, Inc. | jquery.org/license */
```

JedWatson:
@Daniel15 I've added the `!`. I understand it's really important to be able to track the license of included code so that's fine, it's 145 chars which is not a big number. I'll publish a shorter URL for the license too soon which will limit it further.

PR #32, opened 2015-05-04
add travis configs

longlho:
@dcousens @JedWatson separate out travis CI sauce


dcousens:
Thanks @longlho

JedWatson:
I just enabled this in Travis, so we should start seeing builds for commits now. Thanks @longlho and @dcousens :)

longlho:
noice

PR #36, opened 2015-05-12
speedup

silentroach:
Strings concatenation is slow, here is simple speedup

    * local#strings x 6,304,350 ops/sec ±0.72% (95 runs sampled)
    *   npm#strings x 3,577,208 ops/sec ±1.38% (93 runs sampled)

iojs v2.0.1

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,7 +7,7 @@
 function classNames () {
 	'use strict';
 
-	var classes = '';
+	var classes = [];
 
 	for (var i = 0; i < arguments.length; i++) {
 		var arg = arguments[i];
@@ -16,21 +16,21 @@ function classNames () {
 		var argType = typeof arg;
 
 		if ('string' === argType || 'number' === argType) {
-			classes += ' ' + arg;
+			classes.push(arg);
 
 		} else if (Array.isArray(arg)) {
-			classes += ' ' + classNames.apply(null, arg);
+			classes.push(classNames.apply(null, arg));
 
 		} else if ('object' === argType) {
 			for (var key in arg) {
 				if (arg.hasOwnProperty(key) && arg[key]) {
-					classes += ' ' + key;
+					classes.push(key);
 				}
 			}
 		}
 	}
 
-	return classes.substr(1);
+	return classes.join(' ');
 }
 
 // safely export classNames for node / browserify


dcousens:
@silentroach interesting,  we showed different results for the same operation.

We should compare results with node 0.12 and various browsers.

silentroach:
node v0.12.2

    * local#strings x 6,936,765 ops/sec ±0.66% (101 runs sampled)
    *   npm#strings x 3,713,862 ops/sec ±1.06% (98 runs sampled)

dcousens:
@silentroach why the close?

silentroach:
seems like package is not supported any more

JedWatson:
@silentroach ?

dcousens:
This speed up is consistent with my own results,  however,  I'm still not sold on the results themselves.

The `mix` scenario is the most likely reason someone would use `classnames`, and it has a measurable slow down.  The pure `strings` scenario is actually the least likely, as it implies you have all your data as strings to begin with,  with no conditionals.

I think this performance will change so drastically, we would probably need a better benchmarking suite before we merged this.

dcousens:
Thanks @silentroach but in light of the above,  we'll opt not to merge this.  If you have some more benchmarks or if this changes as we time goes on,  please don't hesitate to make another PR.

impinball:
@dcousens IMHO a speedup is a speedup. I would take a speed boost PR to even cold paths if it weren't extremely hackish.

Also, normally in this case for concatenation, I would use an array + final `.join(' ')` instead of string concatenation, as that's how my brain thinks. I think of the class list as a space-delimited list, and I would code it accordingly.

dcousens:
@impinball no doubt,  it is a space delimited list.
However, we have more than 1 use case to account for here,  and there was a slow down in the other use cases.
It'd be great if we had some statistics on which use cases were the most popular,  but without that,  I can only assume we shouldn't allow for any performance regressions like in the above.

impinball:
@dcousens 

I think the likely common case would be strings and objects, with a single object being most common. I wonder if [this (see JSPerf code)](https://jsperf.com/custom-hashmap-implementation-vs-for-in-vs-for-loop/5) would fare better? I'll see what I can do with dedupe.js in a bit.

dcousens:
Sounds great @impinball!
A JSPerf and/or results from the localized `benchmarks/` would be really useful to be included in any resultant PR's.

dcousens:
Moved discussion to #50 :)

PR #38, opened 2015-05-28
Add umd wrapper

### LINKED ISSUE #37: 'use strict' and eval
I'm having this specific problem using `classNames` with `webpack` `devtool` option set to `eval`. It only happens on Chrome and when the console is not open. The issue is calling `classNames` inside the `classNames` function with the `'use strict'` statement inside the function (https://github.com/JedWatson/classnames/blob/master/index.js#L22), this will cause a `ReferenceError`.

See: 

https://github.com/webpack/webpack/issues/417

Explanation:

https://github.com/ftlabs/fastclick/pull/270

I would be happy to submit a PR with a umd wrapper if you guys are ok with it :)

Thanks for the great lib!

daviferreira:
Fixes #37 

Just in case removing `use strict` affects performance. I ran the benchmark and didn't noticed much of a difference. Not sure if I need to add tests to this either. And let me know if you guys want a jsperf.

Cheers!

### DIFF
--- a/dedupe.js
+++ b/dedupe.js
@@ -4,91 +4,92 @@
   http://jedwatson.github.io/classnames
 */
 
-var classNames = (function () {
+(function () {
 	'use strict';
 
-	function _parseArray (resultSet, array) {
-		var length = array.length;
+	var classNames = (function () {
+		function _parseArray (resultSet, array) {
+			var length = array.length;
 
-		for (var i = 0; i < length; ++i) {
-			_parse(resultSet, array[i]);
+			for (var i = 0; i < length; ++i) {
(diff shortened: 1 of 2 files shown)


dcousens:
@daviferreira would it be an issue if we wrapped this as `require("classnames/umd")`?

JedWatson:
Nice @daviferreira, really clean. I'm happy to merge this. Thanks.

(also - I _knew_ I recognised you - the medium editor component, been following that for a while. nice work!)

vlkosinov:
thank you! it's realy helped me to load this module with requirejs

Issue #40, opened 2015-06-08
Broke if you both use AMD and commonjs on the same page


kevinsimper:
The latest release if you use AMD and commonjs/browserify on the same page.

dcousens:
Can't we all jut get along.

@JedWatson can we just export like normal, but then just recommend people to use a wrapper that does it properly for them?
In the end, this is a node package.

IMHO, anyway.

JedWatson:
@kevinsimper can you please clarify what you mean by "broke"? and how it _should_ behave in that case?

Looks to me like fixing this might be as simple as putting this condition first:

```
if (typeof module !== 'undefined' && module.exports)
```

daviferreira:
I'm also curious about what `broke` means here, this is just a very common wrapper, used everywhere. And I'm also curious about the scenario of having both amd and commonjs on the same page.

Cheers!

kevinsimper:
If you are on a webpage where you both have requirejs and use browserify, then the module.exports will never run and therefore the browserify version will not work.

Because both amd and module can be defined on the same page. So the `else if` prohibites the use of both.

I can agree that it is stupid to use both on the same page, but sometimes its something you can not decide.

JedWatson:
@kevinsimper can you confirm whether my proposed change would work, or would you expect it to expose _both_ CommonJS and AMD? A PR that you've confirmed meets your requirements would be welcome.

dcocchia:
I'm seeing this issue when using react-boostrap, which includes classnames as a dependency. I'm also using browserify and babelify to bundle.

<code>Uncaught Error: Mismatched anonymous define() module: function () {
			return classNames;
		}
</code>

This is fixed by flipping the conditions, as @JedWatson suggests:

<code>
	if (typeof module !== 'undefined' && module.exports) {
		module.exports = classNames;
	} else if (typeof define === 'function' && typeof define.amd === 'object' && define.amd) {
		// AMD. Register as an anonymous module.
		define(function () {
			return classNames;
		});
	} else {
		window.classNames = classNames;
	}
</code>

rickychien:
I ran into the same problem after #41 merged. My use case is using requirejs but it also provide module.export at 
https://github.com/JedWatson/classnames/blob/master/index.js#L38

therefore, it doesn't execute second conditional statement for this case.

daviferreira:
Sorry, @JedWatson :heart:

yordis:
@JedWatson is this fixed and published?

[...]

PR #41, opened 2015-06-28
fixes issue where AMD and CommonJS are both used on the same page.

dcocchia:
This is to fix issue talked about here: https://github.com/JedWatson/classnames/issues/40


JedWatson:
Thanks @dcocchia

PR #47, opened 2015-09-28
Name the AMD module.

impinball:


dcousens:
Why?All

JedWatson:
@impinball sorry Dan's terse response didn't go down well with you, FWIW there's assumed knowledge about the value of this that's not explained above.

I'd be happy to merge this change if you want to reopen the PR.

impinball:
Okay...I'll actually explain this one (and not be such an asshole like I was):

If it's named, it's like most other libraries in that if it's loaded directly in the browser, you can use it just like, say, jQuery. It can be imported with just the following:

```js
define(["classnames"], function (classNames) {
  var style = classNames("foo", "bar")
  // ...
})
```

JedWatson:
Thanks @impinball! I'm getting a conflict on this PR, would you mind resolving that so I can merge this?

impinball:
@JedWatson I'm having trouble finding the conflict. I'll just open another PR. This one's simple enough to do.

impinball:
See #47.

PR #48, opened 2015-09-28
Safer `hasOwnProperty` check

impinball:
Just in case someone has a weird and terrible `hasOwnProperty` class name.

### DIFF
--- a/dedupe.js
+++ b/dedupe.js
@@ -6,6 +6,8 @@
 
 (function () {
 	'use strict';
+	
+	var hasOwn = {}.hasOwnProperty
 
 	var classNames = (function () {
 		function _parseArray (resultSet, array) {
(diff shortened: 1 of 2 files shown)


dcousens:
IMHO this isn't worth it,  this module wasn't designed to deal with complex objects.
I feel like this would have come up sooner if people were actually using it like this.

NACK,  but its a trivial change so I'm not bothered.

dcousens:
IMHO this isn't worth it,  this module wasn't designed to deal with complex objects.
I feel like this would have come up sooner if people were actually using it like this.

NACK,  but its a trivial change so I'm not *that* bothered.

dcousens:
Thanks for your enlightening comments @impinball!
Having now actually read the code changes on GitHub instead of in the dark on my phone at 2am,  I realise this change already existed in the code base!
You merely aliased the `hasOwnProperty` to avoid the look up at run time.

If you want to re-open your PR I'll merge it,  otherwise,  I'll merge them in later today.

impinball:
@JedWatson #49

impinball:
Also, I don't get why #36 attracted disagreement.

PR #49, opened 2015-09-29
Safer `hasOwnProperty` check

impinball:


dcousens:
@impinball could you split the polyfill to another PR?
It may be a breaking change for some unfortunately so we'd have to bump major to adhere to SEMVER, so I'd prefer to separate the discussion.

dcousens:
Wait nevermind,  there is already a polyfill requirement for <= IE8.

dcousens:
@impinball `Object.keys` may be slower,  so I'll have to bench it.

impinball:
@dcousens Slower if you're using the polyfill. It's actually implemented significantly faster in every other case for objects in both V8 and SpiderMonkey.

impinball:
It is slower in SpiderMonkey, though. But it simplified things.

dcousens:
For others,  this PR was merged as "safer" as shown in a comment by @impinball:

> object.hasOwnProperty(prop) > {}.hasOwnProperty.call(object, prop) somehow...Not smart if someone uses hasOwnProperty as a class name. Even worse if user input isn't sanitized first, and parsed into a class name. Oops.

Ref: https://github.com/lhorie/mithril.js/issues/545#issuecomment-144072682

impinball:
Thanks for cross posting. I didn't intend to sound like an asshole (although I know I did).

dcousens:
@impinball

> It is slower in SpiderMonkey, though

Reference / benchmark results?

> It's actually implemented significantly faster in every other case for objects in V8.

Any reference for that?  Just curious.

impinball:
@dcousens It is slower in the Firefox browser I tested it right before that post via JSPerf. Although in practice, it's not really that much of a difference.

PR #50, opened 2015-09-30
use [].join over concatenation

dcousens:
A rebase of #36.

Using node `v4.1.1`:

``` javascript
* local#strings x 6,045,951 ops/sec ±1.58% (97 runs sampled)
*   npm#strings x 3,216,399 ops/sec ±4.18% (88 runs sampled)

> Fastest is local#strings

* local#object x 2,542,831 ops/sec ±1.06% (98 runs sampled)
*   npm#object x 2,526,207 ops/sec ±1.12% (91 runs sampled)

> Fastest is local#object |   npm#object

* local#mix x 948,129 ops/sec ±1.53% (93 runs sampled)
*   npm#mix x 1,265,072 ops/sec ±0.97% (96 runs sampled)

> Fastest is npm#mix

* local#arrays x 623,087 ops/sec ±0.69% (98 runs sampled)
*   npm#arrays x 814,812 ops/sec ±0.69% (103 runs sampled)

> Fastest is npm#arrays
```

Performance TL;DR:
* pure string case is 180% faster
* objects are the same
* (string, object), arrays and multiple objects (mix) are all 25% slower

Its a trade off,  I'd wager the two most important cases are the pure string and (string, object) cases,  in which case I'm still not sure what the decision should be.

Perhaps some results from other engines could be the deciding factor?  @impinball

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,7 +11,7 @@
 	var hasOwn = {}.hasOwnProperty;
 
 	function classNames () {
-		var classes = '';
+		var classes = [];
 
 		for (var i = 0; i < arguments.length; i++) {
 			var arg = arguments[i];
@@ -20,19 +20,19 @@
 			var argType = typeof arg;
 
 			if (argType === 'string' || argType === 'number') {
-				classes += ' ' + arg;
+				classes.push(arg);
 			} else if (Array.isArray(arg)) {
-				classes += ' ' + classNames.apply(null, arg);
+				classes.push(classNames.apply(null, arg));
 			} else if (argType === 'object') {
 				for (var key in arg) {
 					if (hasOwn.call(arg, key) && arg[key]) {
-						classes += ' ' + key;
+						classes.push(key);
 					}
 				}
 			}
 		}
 
-		return classes.substr(1);
+		return classes.join(' ');
 	}
 
 	if (typeof module !== 'undefined' && module.exports) {


JedWatson:
To add some more background to the history of this; the first version of this package used array concatenation w/ `.join(' ')`; @chenglou optimised it early on to use string concatenation instead because we found it to be more performant in most cases. That may have changed with updates to V8 since then, but keep in mind the primary environment for this package is browsers, so for changes like this we really need to be looking at JSPerf results and not relying on our benchmarks alone.

>  I would use an array + final .join(' ') instead of string concatenation, as that's how my brain thinks (ref #36)

I agree, thus the original implementation, however we make a lot more concessions for performance in this library than in most and are very conservative about merging changes that negatively impact the performance of some use-cases in favour of others.

FWIW the most common use-case I've seen in the real world is, as @impinball said in #36, a mix of strings and objects, or a single object. No easy way to validate that though without doing code analysis on some of the dependent packages, which might be a valuable exercise if someone has time to do it (I believe @jdalton did something like this recently for lodash)

jdalton:
Naw the old array vs string issue is a classic misread micro bench. Summoning @mraleph to lay it down.

JedWatson:
@jdalton thanks for jumping in :)

dcousens:
Depending on this discussion,  we'll revert https://github.com/JedWatson/classnames/commit/3ff23cabbebe2c9ae2bba5be40eb4af9f74e20d5 (but keep the `hasOwnProperty` change, for safety).

impinball:
@jdalton Thanks for that little bit of input!
@dcousens #51 - commit was easier than revert here.

mraleph:
@jdalton I don't think it's a clear cut. join over array of strings can be faster than building a tree of cons-strings and then flattening it in `substr(1)`. 

There are cases where concat is faster, there are cases where join is - depends on many factors.

jdalton:
String concats can leverage deferred flattening where arrays can't. I wouldn't advocate using Array#join other than for style preference and not for any perf thing. I can tell you from at least a Chakra engine perspective we've focused on string concats but the array join case is not on our radar as it's a misdirection/hack for a failing of the engine to optimize the intuitive use.

mraleph:
In this particular case you hit `substr(1)` so delayed flattening is unlikely. Also it depends on the structure of this cons string you flatten. Here the structure is a bit hairy because code does `x += ' ' + y` - which will cost more to flatten than `x += y`. 

We can crunch some numbers in terms of amount/total size of temporary objects allocated - and the amount of times strings is copied around - I think array based concatenation will allocate less given that the number of classnames is small, backing storage for the array probably grows only once.

jdalton:
Naw, this is a micro-opt through and through. This is a DOM className util where both get millions of operations a second which is well over the average usage of a lib like this.

mraleph:
Sure it's a micro-opt. You wanted me to lay it down, I am laying it down - the fact that `join` can be faster is not simply a misinterpretation - it can. Likewise concatenation can also be faster in certain cases.

Whether "faster" actually matters in a particular context depends on the context - here probably not.

impinball:
The main reason I use it is more for style reasons, anyways.

In terms of speed, IIRC, it's faster to build strings through the concatenation operator, and then at the end, do `str.concat()` for faster string parsing later on. The V8 source does that itself in multiple places (A CL of mine ran into that very discussion). I suspect other engines are similar.

Another thing is that JS strings in V8 are internally implemented as a persistent trie of string slices, if I understand correctly.

dcousens:
So,  in conclusion,  I'm thinking we take the readability benefit and slight performance increase of the `[].join`?

impinball:
If you want. It's about the same. Not arguing the color of the bikeshed here.

dcousens:
@JedWatson so, we merging?

dcousens:
@JedWatson merge it.  Its probably worth it.

JedWatson:
@dcousens ok.

PR #51, opened 2015-09-30
Remove Object.keys implementation (for now).

impinball:
See https://github.com/JedWatson/classnames/pull/50#issuecomment-144226892


JedWatson:
oops - in the flurry of comments and updates earlier I missed this :grimacing: 

publishing a patch now, thanks @impinball

PR #55, opened 2015-09-30
fix(readme): add semver informations, fixes #53

FGRibreau:


dcousens:
Personally I don't think this is necessary.
Any package on `npm` with a semantic version in its `package.json` is supposed to be following SEMVER.
If you feel that this actually exceptional,  then maybe a 1 liner with a link to http://semver.org/ would be appropriate.

JedWatson:
There are still to many packages on npm that don't follow semver, and without an explicit statement it's impossible to know. So if this helps build trust and adoption (or reduce stress for users when updates are released) I'm happy to include it.

Maybe something slightly more subtle though? Like just a "This package follows the [semver](http://semver.org/) standard of versioning." note.

dcousens:
> If you feel that this is actually exceptional, then maybe a 1 liner with a link to http://semver.org/ would be appropriate.

>  Maybe something slightly more subtle though? Like just a "This package follows the semver standard of versioning." note.

:+1:

FGRibreau:
> Any package on npm with a semantic version in its package.json is supposed to be following SEMVER.

This is so wrong, I could recite a lot of packages that don't follow it

> There are still to many packages on npm that don't follow semver, and without an explicit statement it's impossible to know

:+1:

JedWatson:
@FGRibreau thoughts on my edit / one-liner?

FGRibreau:
@JedWatson indeed, sounds good to me :)

JedWatson:
@FGRibreau I (hopefully) cleared up some other recent concerns too :)

Let me know if you've got any feedback on that update.

dcousens:
> This is so wrong, I could list a lot of packages that don't follow it

I said *supposed*,  although I'm still curious what packages don't if you had a few in mind.

PR #60, opened 2015-11-08
Update README.md to properly reflect working API

diegolacarta:
import from 'classnames/bind' and assigned the binded function to a new variable (babel reporting classNames is readonly)


JedWatson:
Aaah yes, thank you :)

PR #61, opened 2015-11-25
Add deps parameter to the AMD module

flipperkid:
Hey this will avoid an issue when using Dojo loader: TypeError: deps.map is not a function

http://stackoverflow.com/a/29312237/5605390


dcousens:
LGTM?

JedWatson:
Thanks @flipperkid, looks good!

PR #64, opened 2015-12-08
Added falsey example

JohnAGill:
Hey I found the lack of a falsey example pretty confusing, let me know if you have a better one.


dcousens:
Good catch @JohnAGill :+1:

PR #76, opened 2016-04-12
Fix template literal delimiters in README example

barneycarroll:


JedWatson:
Thanks @barneycarroll :)

PR #77, opened 2016-04-21
Prefix numeric classnames with underscore

igl:
The naming rule for css classes is `-?[_a-zA-Z]+[_a-zA-Z0-9-]*`
As a convenience this will prefix a number with `_` instead of throwing or logging a warning.

### DIFF
--- a/bind.js
+++ b/bind.js
@@ -19,8 +19,10 @@
 
 			var argType = typeof arg;
 
-			if (argType === 'string' || argType === 'number') {
+			if (argType === 'string') {
 				classes.push(this && this[arg] || arg);
+			} else if (argType === 'number') {
+				classes.push(this && this['_' + arg] || '_' + arg);
 			} else if (Array.isArray(arg)) {
 				classes.push(classNames.apply(this, arg));
 			} else if (argType === 'object') {
--- a/dedupe.js
+++ b/dedupe.js
@@ -20,7 +20,7 @@
 		var hasOwn = {}.hasOwnProperty;
 
 		function _parseNumber (resultSet, num) {
-			resultSet[num] = true;
+			resultSet['_' + num] = true;
 		}
 
 		function _parseObject (resultSet, object) {
--- a/index.js
+++ b/index.js
@@ -19,8 +19,10 @@
 
 			var argType = typeof arg;
 
-			if (argType === 'string' || argType === 'number') {
+			if (argType === 'string') {
 				classes.push(arg);
+			} else if (argType === 'number') {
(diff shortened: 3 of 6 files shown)


dcousens:
I'm not against this,  but two things come to mind:

* Compatibility
* Performance

Will this break expectations people might have had with this module?  Even if the behaviour is erroneous in its final form,  people might be doing post-processing of the resultant string which this would now break.  So,  potential major version bump required.

Also,  does this impact on performance in any way?
This module typically gets used *over and over*,  I think its pertinent to verify you aren't changing anything that might affect that.

igl:
Valid points. I don't see any performance impact since this is only affecting numbers which you can not really use right now (correctly). If anything the string resolution should be faster because of the shortened if-case. I would consider reordering those if switches since Strings and Numbers are what i would least likely use this module for and could be put at the bottom.

danbovey:
Even though it's called classNames, it really has nothing to do with CSS, it's dealing with the DOM and HTML attributes. I think it's out of the scope of this package for it to worry about whether classes given to it are actually syntactically correct.

I'd want a warning telling me, hey you!... learn CSS!

pygy:
I agree with Dan here, there are many ways a class can be wrong. You could alternatively provide a wrapper that validates the output.

```JS
// validate.js
var validator  = /(?:^|\s+)-?[_a-zA-Z]+[-\w]*|(.)/g

function validate(classnames) {
  return function() {
    var res = classnames.apply(this, arguments)

    var match
    while (match = validator.exec(res)) {
      if (match[1]) {
        validator.lastIndex = 0 // rewind the regexp
        throw new Error(
          "Unexpected character "+ JSON.stringify(match[1]) +
          " at position " + match[2] +
          " in class list\n" + JSON.stringify(res)
        )
      }
    }
  }
}
```

Then

```JS
var classnames = require("classnames/validate")(require("classnames"))
// use it as usual.
```

It is also compatible with the `bind` version, pre- and post-binding.

Would you mind a PR?

Note that the code above is untested and may be subtly wrong, I'll give it a second look and add tests if you agree with the idea.

dcousens:
NACK but performance reasons, I don't think this is our responsibility.  If you think otherwise,  make a PR with it under a separate file,  `/safe` maybe.

dcousens:
I don't know,  I'm going to defer my opinion on this to @JedWatson

igl:
Coming back to this now... I don't think this was a clever idea anymore.

- Not worth the breakage. classNames is too widely used.
- Weird unpredictable behaviour.

PR #85, opened 2016-06-09
travis: drop iojs, add stable node

avindra:
Swap out io.js testing for stable node


dcousens:
`"stable"`
Who knew.
Maybe throw in 4, 5 and 6 for good measure and we'll call it a deal

avindra:
Updated PR

JedWatson:
:+1: thanks @avindra!

PR #86, opened 2016-06-17
Update LICENSE to a YEAR-present format

buildbreakdo:
Picked this up from the Facebook React Repo so you don't have to update the year.. well every year! That gets old.

See here:
https://github.com/facebook/react/blob/master/LICENSE

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2016 Jed Watson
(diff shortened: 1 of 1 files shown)


dcousens:
That doesn't look right,  and I suspect something is up with how they are managing that copyright.
It can't just be assumed that the copyright is permanently "present"...

dcousens:
Meh,  this lasts for 70 years.
I don't think it matters much to keep it updated.

PR #90, opened 2016-07-25
test configs: use standard testing pattern

avindra:
Instead of doing `npm run unit`, just configure it in a way that we can run `npm test`. This simplifies both `package.json` and `.travis.yml`


dcousens:
`npm test` would have already worked, but sure

PR #92, opened 2016-07-27
Supports react class with variables css modules resolve

aemoe:
When using react, there will be a problem, plug-in and can not support the use of variables in the object, the authors hope to see, plus the support, thank you!

For example:

If I want use variables

```js
let classNames1 = 'class1';
let classNames2 = 'class2';
let classNames3 = 'class3';

reactClass([{
    classNames1: true,
    classNames2: false,
    classNames3: true
}]) // => 'classNames1 classNames3'
```
so I edit the code,it can use like below:
```js
let classNames1 = 'class1';
let classNames2 = 'class2';
let classNames3 = 'class3';

reactClass([{
    'classname': classNames1,
    'display': true
}, {
    'classname': classNames2,
    'display': false
}, {
    'classname': classNames3,
    'display': true
}]) // => 'class1 class3'
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -24,11 +24,17 @@
 			} else if (Array.isArray(arg)) {
 				classes.push(classNames.apply(null, arg));
 			} else if (argType === 'object') {
-				for (var key in arg) {
-					if (hasOwn.call(arg, key) && arg[key]) {
-						classes.push(key);
-					}
-				}
+				if ('classname' in arg && 'display' in arg) {
+                    if (hasOwn.call(arg, 'display') && arg['classname']) {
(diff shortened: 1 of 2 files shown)


JosephClay:
The first example could be written like this without changing the library:

```js
let classNames1 = 'class1';
let classNames2 = 'class2';
let classNames3 = 'class3';

reactClass([{
    [classNames1]: true,
    [classNames2]: false,
    [classNames3]: true
}]) // => 'class1 class3'
```

aemoe:
thank u very much~

ppo279:
umm,

PR #100, opened 2016-09-26
Added benchmarks for browsers

victor-homyakov:
- Added HTML page
- Refactored benchmarking code to allow its reuse

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,5 @@ node_modules
 
 # Mac OS X DS_Store
 .DS_Store
+
+benchmarks/runInBrowser.bundle.js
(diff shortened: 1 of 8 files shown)


dcousens:
Rather than totally bork all the existing benchmark code with browser boilerplate.  Could you just make a browserify target instead and include that in the page?

dcousens:
ping @victor-homyakov , up to address https://github.com/JedWatson/classnames/pull/100#issuecomment-250365143

victor-homyakov:
@dcousens could you take a look?
- added targets and static HTML page to browserify and run code in browser
- added `setTimeout`s between tests to make code more or less responsive in browser

dcousens:
Thanks @victor-homyakov - awesome :+1:

PR #103, opened 2016-11-09
Add TypeScript declaration.

### LINKED ISSUE #102: Bundle TypeScript definitions
Would you be open to bundling a .d.ts TypeScript definition for this package, so that it can seamlessly be used from TypeScript?

There are existing definitions https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/classnames that could be copied over, so it should be fairly trivial.

bradleyayers:
This resolves #102

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,9 @@
+type ClassValue = string | number | ClassDictionary | ClassArray;
+
+interface ClassDictionary {
+	[id: string]: boolean;
+}
+
+interface ClassArray extends Array<ClassValue> { }
+
+export default function(...classes: ClassValue[]): string;
--- a/index.d.ts
+++ b/index.d.ts
@@ -12,4 +12,4 @@ interface ClassNamesFn {
 
 declare var classNames: ClassNamesFn;
 
-export = classNames
+export default classNames
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,9 +1,15 @@
-type ClassValue = string | number | ClassDictionary | ClassArray;
+declare type ClassValue = string | number | ClassDictionary | ClassArray | undefined | null | false;
 
 interface ClassDictionary {
-	[id: string]: boolean;
+	[id: string]: boolean | undefined | null;
 }
 
 interface ClassArray extends Array<ClassValue> { }
 
-export default function(...classes: ClassValue[]): string;
+interface ClassNamesFn {
+	(...classes: ClassValue[]): string;
+}
+
+declare var classNames: ClassNamesFn;
+
+export = classNames
--- a/bind.d.ts
+++ b/bind.d.ts
@@ -1,0 +1,9 @@
+import * as base from './index';
+
+interface IClassNamesBind {
+	bind: (styles: any) => typeof base;
+}
+
+declare var classNames: IClassNamesBind;
+
+export = classNames;
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "type": "git",
     "url": "https://github.com/JedWatson/classnames.git"
   },
+  "types": "./index.d.ts",
   "scripts": {
     "benchmarks": "node ./benchmarks/run",
     "unit": "mocha tests/*.js",


bradleyayers:
Any further thoughts on this @JedWatson?

cwmacdon:
@dcousens @bradleyayers 

It would be nice if there were definitions for the alternate `classnames/dedupe` and `classnames/bind` versions.

How does this look for `bind.d.ts`?

```
import { default as base } from './index';

interface IClassNamesBind {
    bind: (styles: any) => typeof base;
}

declare var classNames: IClassNamesBind;

export default classNames;
```

bradleyayers:
@cwmacdon looks good to me — I'd love to see that get merged in too 👍.

jedmao:
@bradleyayers is there any reason these definitions [don't match the ones on DefinitelyTyped 100%](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/classnames/classnames.d.ts)?

Also, the [publishing doc]() states the following:

> Also note that if your main declaration file is named index.d.ts and lives at the root of the package (next to index.js) you do not need to mark the "types" property, though it is advisable to do so.

Perhaps you can add it?

```json
{
    "types": "./index.d.ts"
}
```

bradleyayers:
@jedmao I'm not sure I considered just copying over the @types/ code, maybe I wanted to avoid licensing issues, but using it (or matching it) indeed seems more desirable than having a set of deviating definitions.

Having the .d.ts placed next to the index.js should indeed work. I'll give that a go and update the PR

jedmao:
It will work, but they still recommend being explicit about it. I think what i was trying to say before its that if what you have here is better than the one on DT, perhaps you should submit a PR to DT first so that they are in sync. If you do so, submit to the types-2.0 branch.

bradleyayers:
I'll compare and take the best and upstream if necessary. Thanks

cwmacdon:
Any progress on this @bradleyayers @jedmao? Anything I can do?

bradleyayers:
@cwmacdon I haven't had time to get back to this. But what you could do is to look into whether we can just copy the DefinitionTyped definitions into `classnames` directly (I'm not sure if there's licensing issues). This would avoid having two different definitions.

cwmacdon:
@bradleyayers So I tried copying the Definitely Typed version into my `node_modules/classnames/index.d.ts`

Then I wrote a file to use it
```
import classNames from 'classNames';
```

and it gave the error
`File {.../classNames/index.d.ts} is not a module.`

The definition that you have included above works perfectly.

@adidahiya essentially wrote the DefinitelyTyped version, maybe he can weigh in?

adidahiya:
there are no licensing issues. I think you are free to copy the definitions I wrote for DefinitelyTyped, just make sure to add the package to [`notNeededPackages.json`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a8dc31d5a2a55bf458834680d4653bc452949d84/notNeededPackages.json). I'll take a look at the code here now

index.d.ts
```diff
@@ -0,0 +1,9 @@
+type ClassValue = string | number | ClassDictionary | ClassArray;
```
adidahiya: please make this match the DT typings. https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a8dc31d5a2a55bf458834680d4653bc452949d84/classnames/index.d.ts

index.d.ts
```diff
@@ -0,0 +1,9 @@
...
+
+interface ClassDictionary {
+	[id: string]: boolean;
```
adidahiya: same here, this needs to allow `undefined` and `null`. also, personal preference (generally in the TS community), can you use 4-space indentation instead of tabs?

index.d.ts
```diff
@@ -0,0 +1,9 @@
...
+}
+
+interface ClassArray extends Array<ClassValue> { }
+
+export default function(...classes: ClassValue[]): string;
```
adidahiya: most consumers are currently doing `import * as classNames from "classnames";`, and this will break that. actually I don't even think this is the shape of the export; see [this part of the DT readme](https://github.com/DefinitelyTyped/DefinitelyTyped#a-package-uses-export--but-i-prefer-to-use-default-imports-can-i-change-export--to-export-default) for guidance.

```sh
$ node -p 'require("classnames").default'
undefined
$ node -p 'require("classnames")'
[Function: classNames]
```

cwmacdon: @adidahiya There are many ways to do this...
`import classNames from 'classNames';`
`import * as classNames from 'classNames';`
`import classNames = require('classNames');`

Is there a "best" way? Is there a way to write the definition to work with all of the above?

adidahiya: both `import classNames from 'classNames';` and `import * as classNames from 'classNames';` are generally OK. The last one that uses `require()` is discouraged because it is old CommonJS-style syntax.

but, like I said, the DefinitelyTyped README stipulates that you should not declare a default export unless `require("classnames").default` actually returns the module.

cwmacdon: OK Cool good to know.

Looking at your definition on DefinitelyTyped, what is the purpose of the `classnames` module wrapped around the export?

```
declare module "classnames" {
	export = classNames
}
```

cwmacdon:
@bradleyayers Thanks to input from @adidahiya and @jedmao, Could you update the PR with the changes outlined below?


1) `index.d.ts` with the definition from DefinitelyTyped with one slight change: Remove the `declare module "classnames"` wrapper around the export.

2) Add a `bind.d.ts` with following:
```
import * as base from './index';

interface IClassNamesBind {
    bind: (styles: any) => typeof base;
}

declare var classNames: IClassNamesBind;

export = classNames;
```

3) Change `package.json` to explicitly point to the type definitions:
```
    "types": "./index.d.ts"
```

Then once this is complete and merged, a PR to DefinitelyTyped will be needed to add classNames to the [notNeededPackages.json](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a8dc31d5a2a55bf458834680d4653bc452949d84/notNeededPackages.json) to help users know that the types are included in this package now.

index.d.ts
adidahiya: Ah, you don't need that part. That was a remnant from before we had `@types` on NPM -- when you use those typings through `tsd` or the `typings` tool, you would need the ambient declaration of the `"classnames"` module. Now that we have node-style resolution of index.d.ts, that part isn't needed.

cwmacdon: @adidahiya Would you recommend updating the DefinitelyTyped definition to remove it as well? Or should we just leave it out in this PR?

adidahiya: Yeah, updating the DT definition sounds like a good idea now that the repo has switched over to `@types` completely. Once a release of classnames is made that includes definitions though, we should remove those typings from DT (and update notNeededPackages.json in the same PR).

cwmacdon:
@JedWatson Any chance a new release could be cut with this PR?

cwmacdon:
@dcousens Perhaps you could merge this PR? I see you've kind of taken over merging PRs since mid last year.

adidahiya:
bump again here @JedWatson @dcousens, thanks

adidahiya:
This PR just merged in DT so you should be able to pick up the latest `@types/classnames` with most of these improvements. https://github.com/DefinitelyTyped/DefinitelyTyped/pull/16420

cwmacdon:
@adidahiya Thanks for the heads up. I have doubts this will ever get merged in here, so @types/classnames will be the best we can get.

dcousens:
@cwmacdon I was just added as a contributor.
Is this PR up to date?

maciejw:
@bradleyayers can you tell me why you decided to go with export = classNames; instead of export default, I'm curious

this does not work with allowSyntheticDefaultImports = false witch is default setting in ts.

webpack can handle this when I do

```js
import  * as classNames from 'classnames'
```

but rollup will say  

> Cannot call a namespace ('classNames')

with allowSyntheticDefaultImports = true both works fine using this
```js
import  classNames from 'classnames'
```


export = has many issues and it requires this kind of import
```js
import classNames = require('classnames')
```
(https://www.typescriptlang.org/docs/handbook/modules.html)

adidahiya:
@maciejw you probably want the new `--esModuleInterop` flag added in TS 2.7.

> this does not work with allowSyntheticDefaultImports = false witch is default setting in ts.

Note that this only affects type checking, not JS emit.

See the [DefinitelyTyped README](https://github.com/DefinitelyTyped/DefinitelyTyped/#a-package-uses-export--but-i-prefer-to-use-default-imports-can-i-change-export--to-export-default) for info about why `export =` syntax is used here.

maciejw:
indeed I have ts 2.7, but I did not change this flag... this esModuleInterop flag break my postcss modules in jest. witch worries me gravely:) 

now, what I did is copied d.ts locally and declared it as another module and that imported types only
 to a file and reexported it as below. I had to copy d.ts because types are not visible when export = is used I think.

```typescript
import * as ns from 'classnames-ts';

const classNames : ns.ClassNamesFn = require('classnames');

export { classNames };
```

this works fine only with require in all those environments: jest, rollup, webpack

bradleyayers:
@maciejw years ago when I opened this PR I used `export =` at the time this library did not have an ES module export (ES modules are not callable, so `import * as classnames from "classnames"; classNames()` is not valid). The TypeScript method of interop with callable commonjs modules is the `export =` declaration.

However since then the library has added support for ES modules by way of a default export (https://github.com/JedWatson/classnames/commit/915186aaa510e0c6be6c2cf20788aac474790e27). So I recommend this PR be changed to use a default export rather than the callable commonjs module.

maciejw:
@bradleyayers I stumbled upon similar case with not callable export = in invariant https://github.com/zertosh/invariant/issues/32 library... I'm sure not last one ;) to sum it up:
commonjs 'module.export' = d.ts 'export =' witch can be imported in es6 only with import * as x from 'x' with exception to x cannot be callable, on top of this different bundlers/transpiles are adding logic, to solve this incompatibility, is this picture accurate?
whould it be a better solution to add module=index.es.js in package.json as a shim? d.ts then would be complatible with indes.es.js and index.js would be left for commonjs only

jedmao:
Isn't it about time this library drop support for [dead/unmaintained Node.js versions](https://github.com/nodejs/Release#release-schedule)? It seems only [Node.js v0.10 and 0.12 are failing this build](https://travis-ci.org/JedWatson/classnames/builds/340676893).

bradleyayers:
@maciejw I've responded in https://github.com/zertosh/invariant/issues/32#issuecomment-365110342

jedmao:
@dcousens I've rebased https://github.com/JedWatson/classnames/pull/162

ssylvia:
Any idea when this will be merged or is there anything I can do to help get this merged?

I have not been able to get types working this library or by using `@types/classname`.

Using:
```js
import classnames from 'classnames';
```

with `"allowSyntheticDefaultImports": true` in tsconfig.

adidahiya:
@ssylvia you should try enabling both `--allowSyntheticDefaultExports` and `--esModuleInterop`

dcousens:
Wait, these were copied from DefinitelyTyped...

Please add the copyright notice to the top of this file,  it should be enough (though IANAL) as the LICENSE is the same.

lukescott:
Is there a reason `ClassDictionary` is more strict that DefinitelyTyped? Real world example that doesn't work with the stricter version:

```
<div className={classnames("textField", error, {error})}>
```

`error` is a string that contains an error codes, such as `"required"`. If the error is `"required"` then className will be `"textField error required"` while no error would be `"textField"`.

With the stricter types you have to do:

```
<div className={classnames("textField", error, {error: !!error})}>
```

A bit of a nit-pick I guess, but the first one is a tad shorter and the library itself allows `any` as it does a truthy check.

Also it looks like the latest published version to npm does not contain the TypeScript definitions.

theKashey:
Still not released.

FDiskas:
Still :D
That means the project is dead?

JedWatson:
@FDiskas it's released

PR #114, opened 2017-03-03
Exclude bower.json and docs from npm package

jorrit:
Another option would be to add a `files` array to package.json.


dcousens:
ACK,  but I'm not going to merge until @JedWatson adds me as a contributor to the npm package.

ping @JedWatson :stuck_out_tongue:

PR #124, opened 2017-04-09
classnames working with arguments object

lxkuz:
If we will use `classnames` with `arguments` object inside a function we get strange behavior
```javascript
var func = function(a, b, c) { 
  return classnames(arguments)
}

func("a", "b", "c"); // => "0 1 2"

```


lxkuz:
I found a little

PR #131, opened 2017-04-30
Adding join method

### LINKED ISSUE #130: improved css modules method
The way to go about css modules seems to be using the `bind` method.

This takes an object that will be used as a map to lookup the actual className from the result of a css module export.  The way in which it does this is instead of the property key, it will do a lookup against `this` and use the result (or the key if it wasn't found)

This is somewhat limiting as you can't combine multiple sources.  It would be valuable to provide a mechanism to attach multiple objects.  The use case would be using a combination of css frameworks in addition to one's own styles.  Presently multiple functions need to be created and used independently: 

```js
import classNames from 'classnames/bind'
import * as bootstrap from 'bootstrap/less/bootstrap'
import * as fontawesome from 'font-awesome/less/font-awesome'
import styles from './styles.css'

const bs = classNames.bind(bootstrap)
const fa = classNames.bind(fontawesome)
const cx = classNames.bind(styles)

bs('col-xs-12', 'col-xs-6')
fa('fa', 'fa-user')
cx('my-style')
```

It would be cool if this could be simplified into a single call.. off the top of my head, I've called it `join`  and placed it next to `bind`, but this could be named anything and placed anywhere, really...

e.g.:

```js
import classNames from 'classnames/bind'

const cx = classNames.join(bootstrap, fontawesome, styles)

cx('col-xs-12', 'col-sm-6')
cx('fa', 'fa-user')
cx('my-style')
```

tswaters:
A few notes --

I've added both `bind` and `join` to the benchmarks, seems bind was missing.  

I wasn't sure what to do for benchmarks for the npm variant `join` as it doesn't exist in npm now.  I left it out.  Still, gives a comparison with `bind`.

I've added the method as a property on the export of the `bind` file so one will be able to use `classNames.join` instead of bind if desired.  

This uses `Object.assign` so I added a note about using a polyfill for older browsers... also made an assumption about the starting version for that as the next minor bump.

Resolves #130

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,8 +186,37 @@ export default class SubmitButton extends Component {
 
 ```
 
+### Alternate `join` version (for [css-modules](https://github.com/css-modules/css-modules))
+
+If you are using [css-modules](https://github.com/css-modules/css-modules), or a similar approach to abstract class "names" and the real `className` values that are actually output to the DOM, and you have multiple sources for classNames, you may want to use the `join` variant.
+
+_Note that in ES2015 environments, it may be better to use the "dynamic class names" approach documented above._
+
+```js
+var classNames = require('classnames/bind');
+
+var styles1 = {
+  foo: 'abc',
+  bar: 'def',
+  baz: 'xyz'
+};
+
+var styles2 = {
+	qux: 'ghi',
+	quz: 'jkl',
+	quuz: 'uvw',
+}
+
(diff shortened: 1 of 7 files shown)


tswaters:
Shoot.  Turns out node 0.x doesn't support `Object.assign`.  I'll need to think on this.  Stay tuned!

PR #133, opened 2017-05-17
Fix issue with empty arrays being pushed

webdeveloperpr:
Arrays that are a 2nd parameter to  classNames, and arrays that contain empty nested arrays get pushed resulting in spaces at the end of the classNames function output. This PR fixes that issue.

### DIFF
--- a/index.js
+++ b/index.js
@@ -22,11 +22,10 @@
 			if (argType === 'string' || argType === 'number') {
 				classes.push(arg);
 			} else if (Array.isArray(arg) && arg.length) {
-				// if the element inside the array is an empty array do not push it
-				if (Array.isArray(arg[0]) && !arg[0].length) {
-					continue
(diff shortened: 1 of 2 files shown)


dcousens:
@webdeveloperpr thanks!
I simplified the example to avoid extra `Array.isArray` calls

index.js
```diff
@@ -22,11 +22,10 @@
 			if (argType === 'string' || argType === 'number') {
 				classes.push(arg);
 			} else if (Array.isArray(arg) && arg.length) {
-				// if the element inside the array is an empty array do not push it
-				if (Array.isArray(arg[0]) && !arg[0].length) {
-					continue
+				var inner = classNames.apply(null, arg);
```
webdeveloperpr: ahh , interesting I didn't think about this.

PR #135, opened 2017-06-19
Update Readme for npm ^5 --save

morajabi:
From npm version 5, there is no need for `--save` flag because it automatically saves package into dependency section in package.json.


newyork-anthonyng:
Does doing `npm install classnames --save` still work in npm version 5?

If it does, I don't think we should add this change. It adds unnecessary complexity to the documentation.

PR #88, opened 2016-07-04
Improve performance in 2 times

alekseykulikov:
**How**:
- use string intead of array
- cache arguments.length

**Benchmark**:
```
$ node -v
v6.2.2

$ npm run benchmarks

> classnames@2.2.5 benchmarks /Users/aleksey/code/experiments/classnames
> node ./benchmarks/run

* local#strings x 1,716,748 ops/sec ±10.68% (80 runs sampled)
*   npm#strings x 804,827 ops/sec ±1.05% (92 runs sampled)
* local/dedupe#strings x 291,698 ops/sec ±3.52% (87 runs sampled)
*   npm/dedupe#strings x 300,405 ops/sec ±6.53% (87 runs sampled)

> Fastest is local#strings

* local#object x 1,278,203 ops/sec ±4.31% (90 runs sampled)
*   npm#object x 588,600 ops/sec ±1.58% (88 runs sampled)
* local/dedupe#object x 472,073 ops/sec ±0.66% (96 runs sampled)
*   npm/dedupe#object x 450,313 ops/sec ±1.60% (93 runs sampled)

> Fastest is local#object

* local#strings, object x 1,230,869 ops/sec ±0.89% (95 runs sampled)
*   npm#strings, object x 583,617 ops/sec ±2.05% (92 runs sampled)
* local/dedupe#strings, object x 254,245 ops/sec ±0.92% (95 runs sampled)
*   npm/dedupe#strings, object x 211,184 ops/sec ±7.14% (81 runs sampled)

> Fastest is local#strings, object

* local#mix x 508,282 ops/sec ±11.10% (74 runs sampled)
*   npm#mix x 346,792 ops/sec ±9.73% (80 runs sampled)
* local/dedupe#mix x 59,928 ops/sec ±16.01% (72 runs sampled)
*   npm/dedupe#mix x 44,578 ops/sec ±14.48% (57 runs sampled)

> Fastest is local#mix

* local#arrays x 321,436 ops/sec ±9.90% (91 runs sampled)
*   npm#arrays x 125,608 ops/sec ±8.53% (75 runs sampled)
* local/dedupe#arrays x 138,630 ops/sec ±3.89% (95 runs sampled)
*   npm/dedupe#arrays x 103,878 ops/sec ±9.44% (72 runs sampled)

> Fastest is local#arrays
```

PR #140, opened 2017-09-14
Use strings for perf bump

gamb:
Results in:

```
* local#strings x 4,714,423 ops/sec ±2.16% (91 runs sampled)
*   npm#strings x 3,817,184 ops/sec ±1.63% (91 runs sampled)
* local/dedupe#strings x 1,661,997 ops/sec ±2.11% (95 runs sampled)
*   npm/dedupe#strings x 1,799,861 ops/sec ±1.44% (94 runs sampled)

> Fastest is local#strings

* local#object x 5,399,003 ops/sec ±1.95% (93 runs sampled)
*   npm#object x 3,067,269 ops/sec ±2.94% (89 runs sampled)
* local/dedupe#object x 2,223,237 ops/sec ±3.09% (91 runs sampled)
*   npm/dedupe#object x 2,088,421 ops/sec ±3.31% (81 runs sampled)

> Fastest is local#object

* local#strings, object x 3,192,509 ops/sec ±4.58% (87 runs sampled)
*   npm#strings, object x 2,909,568 ops/sec ±2.79% (89 runs sampled)
* local/dedupe#strings, object x 1,420,670 ops/sec ±3.99% (84 runs sampled)
*   npm/dedupe#strings, object x 1,553,730 ops/sec ±2.26% (92 runs sampled)

> Fastest is local#strings, object

* local#mix x 2,237,207 ops/sec ±3.26% (89 runs sampled)
*   npm#mix x 2,151,091 ops/sec ±3.75% (94 runs sampled)
* local/dedupe#mix x 735,114 ops/sec ±1.36% (98 runs sampled)
*   npm/dedupe#mix x 736,473 ops/sec ±1.13% (97 runs sampled)

> Fastest is local#mix

* local#arrays x 1,185,741 ops/sec ±2.52% (89 runs sampled)
*   npm#arrays x 955,967 ops/sec ±1.36% (97 runs sampled)
* local/dedupe#arrays x 879,370 ops/sec ±1.56% (97 runs sampled)
*   npm/dedupe#arrays x 852,658 ops/sec ±1.71% (91 runs sampled)

> Fastest is local#arrays
```


newyork-anthonyng:
If you look at this [commit](https://github.com/JedWatson/classnames/commit/0de1ea13fdfce78299e388449d2243f82a0eeac4#diff-168726dbe96b3ce427e7fedce31bb0bc), it looks like we previously used `strings` instead of `arrays`.

@dcousens Do you remember what the rationale was?

dcousens:
@newyork-anthonyng performance yo-yos between versions.

https://github.com/JedWatson/classnames/pull/88
https://github.com/JedWatson/classnames/pull/50
https://github.com/JedWatson/classnames/pull/36

If we can show that there is a speed-up in three popular browsers, and the last 4 node versions,  and cross-verified by at least 3 platforms... I'll merge.

Otherwise,  meh.

gamb:
Ah, the previous implementation used `.substr(1);` on the string at the end. I tried that too, but the perf gain wasn't significant compared to converting an `Array -> String` as you do now.

`String +=` alone is *much* faster by comparison.

But also, meh.

dcousens:
Can you set up a jsperf?

gamb:
I tried to setup a jsperf quite a few times, but it keeps crashing today.

Take a simple implementation:

```javascript
// classNames#str
function classNamesStr(def) {
    var filtered = ""
    for (var key in def) {
        if (def[key] && def.hasOwnProperty(key)) {
            filtered += (filtered ? " " : "") + key
        }
    }
    return filtered
}

// classNames#array
function classNamesAr(def) {
    var filtered = []
    for (var key in def) {
        if (def[key] && def.hasOwnProperty(key)) {
            filtered.push(key)
        }
    }
    return filtered.join(" ")
}

// classNames#substr
function classNamesSubstr(def) {
    var filtered = ""
    for (var key in def) {
        if (def[key] && def.hasOwnProperty(key)) {
            filtered += " " + key
        }
    }
    return filtered.substr(1)
}
```

You should find `classNames#substr` and `classNames#array` are comparable, whereas `classNames#string` is significantly (~ 2x) faster.

For my own purposes I think I'm just going to roll my own function (classNames#string), but feel free to accept/reject this pull :)

gamb:
FWIW here is what I've ended up doing: https://gist.github.com/gamb/f49a06822b90d5545d073df5f401a022

Configurable via factory without loss of perf.

Issue #146, opened 2017-11-07
Render no classnames at all if no valid args passed


glennreyes:
I think, if we pass anything like:
```js
classNames()
classNames({})
classNames(null)
classNames(false)
classNames({ [styles.Something]: false })
```

... it shouldn't render the `class` attribute at all. At the moment it does:
![image](https://user-images.githubusercontent.com/5080854/32493127-0f664954-c3bd-11e7-97b7-1468b80c0279.png)

We often have the case, where we just need one class name based on condition. Of course, a simple ternary would do that too, but we want to keep the codebase consistent by using this lib across the project since it's awesome 💓 .

Why is rendering the empty class attribute in such cases reasonable?

newyork-anthonyng:
Hi @glennreyes,

Take a look at this [issue](See this `issue` [here](https://github.com/JedWatson/classnames/issues/115). Someone wanted to do something similar, but return `undefined` instead.

This might not have backward compatibility issues, and would require a `sub-module`.

glennreyes:
Thanks for pointing to the issue @newyork-anthonyng! [This](https://github.com/JedWatson/classnames/issues/115#issuecomment-284122686) sounds reasonable, so will close it for now and stick for [this](https://github.com/JedWatson/classnames/issues/115#issuecomment-284296632) workaround.

PR #147, opened 2017-11-07
Render nothing at all if no class converted

glennreyes:
#146


glennreyes:
https://github.com/JedWatson/classnames/issues/146#issuecomment-342752294

PR #148, opened 2017-11-13
Return null when classes is an empty array

danhuang1202:
If the ```classes``` is an empty array, the method ```join(' ')``` will return empty string.
But the empty string of react's ```className``` would cause react generate class attribute without value.
So I think it's better to return null when the ```classes``` is an empty array.

### DIFF
--- a/index.js
+++ b/index.js
@@ -35,7 +35,7 @@
 			}
 		}
 
-		return classes.join(' ');
+		return classes.length ? classes.join(' ') : null;
 	}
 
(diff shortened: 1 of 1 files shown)


newyork-anthonyng:
Hi @danhuang1202,

Take a look at this [issue](https://github.com/JedWatson/classnames/issues/146).

danhuang1202:
@newyork-anthonyng, thank you for pointing out this. I will close this pr.

PR #156, opened 2018-03-18
Update React docs to reflect ES6 syntax + add example

erikshestopal:
## Changes
1. I've updated the documentation to better reflect the syntax of modern ES6 since `React.createClass` has been deprecated in favor of using stateless functions or ES6 classes. 
2. I've also added another example to the `Dynamic class names with ES2015` section to illustrate the possibility of using template literals without actually having to use dynamic properties if the condition is always  set to`true`. Saves a few characters of typing in most cases. 
3. Fixed a small typo: `'btn-over'` => `'btn-hover'`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ var arr = ['b', { c: true, d: false }];
 classNames('a', arr); // => 'a b c'
 ```
 
-### Dynamic class names with ES2015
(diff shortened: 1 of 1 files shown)


erikshestopal:
@JedWatson Can this PR get reviewed?

PR #160, opened 2018-04-12
PHP Port

mcharytoniuk:
Hello, I really like this library and I really missed classnames in PHP. I made a port because I wanted to contribute to PHP community also. I would be thankful if you agree to leave a mention in the readme.

Best wishes


samuelbarrionuevo (approved):

dcousens:
Thanks for sharing the interesting library @mcharytoniuk  - in this instance I'm opting not to add the link to the README to not introduce a maintenance burden of keeping that list relevant and up-to-date.

PR #166, opened 2018-06-15
Make the AMD module check less restrictive

alexlafroscia:
If you look at how UMD recommends defining an AMD module, it doesn't
include a check that `define.amd` is an `object`.

https://github.com/umdjs/umd/blob/36fd1135ba44e758c7371e7af72295acdebce010/templates/amdWeb.js#L19

That check was breaking import from the `ember-cli` which supports AMD
modules but does not include `define.amd` as an `object`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: node_js
-before_install:
-  - "npm install npm -g"
+
 node_js:
   - "stable"
   - "4"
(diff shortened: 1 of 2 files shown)


alexlafroscia:
Tests are failing because of this line in your `.travis.yml` file:

https://github.com/JedWatson/classnames/blob/30bd487050de10a016ed2abfd6e602619d166ddc/.travis.yml#L3

It's installing a version of `nom` that isn't actually compatible with many of the versions of Node being tested against. I don't think it's necessary to re-install the `npm` package during your tests, Travis ships with one.

alexlafroscia:
Since the Rollup config there exports a UMD module, it looks like that would also solve my problem!

PR #167, opened 2018-06-21
add __esModule: true when exports

doxiaodong:
since babel _interopRequireWildcard doesn't do the right copy to the new newObj
when the module.exports is a function

and classnames is actually a function exports

see https://github.com/babel/babel/issues/3956
and https://github.com/babel/babel/blob/master/packages/babel-helpers/src/helpers.js#L544

### DIFF
--- a/bind.js
+++ b/bind.js
@@ -38,6 +38,7 @@
 	if (typeof module !== 'undefined' && module.exports) {
 		classNames.default = classNames;
 		module.exports = classNames;
+		module.exports.__esModule = true;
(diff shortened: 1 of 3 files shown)


doxiaodong:
By the way, the reason I use `import * as classnames from 'classnames'` rather than `default import` is https://github.com/mradamczyk/DefinitelyTyped/blob/fae50ad424116d7179e9e9d20651ba17ad229d84/types/classnames/index.d.ts#L24

PR #175, opened 2018-09-07
Use Set for dedupe

### LINKED ISSUE #173: Use `Set` objects for dedupe
Has there been any research done in improving the performance of `dedupe` by utilising a [`Set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set)?

Sets are said to be faster<sup>[[1]](https://github.com/anvaka/set-vs-object) [[2]](https://stackoverflow.com/questions/39007637/javascript-set-vs-array-performance)</sup> and are more idiomatic when dealing with lists of unique elements.

newyork-anthonyng:
See #173 for benchmark numbers.

### DIFF
--- a/tests/dedupe.js
+++ b/tests/dedupe.js
@@ -31,7 +31,7 @@ describe('dedupe', function () {
 	});
 
 	it('joins arrays of class names and ignore falsy values', function () {
-		assert.equal(dedupe('a', 0, null, undefined, true, 1, 'b'), '1 a b');
+		assert.equal(dedupe('a', 0, null, undefined, true, 1, 'b'), 'a 1 b');
 	});
 
 	it('supports heterogenous arguments', function () {
--- a/dedupe.js
+++ b/dedupe.js
@@ -9,11 +9,6 @@
 	'use strict';
 
 	var classNames = (function () {
-		// don't inherit from Object so we can skip hasOwnProperty check later
-		// http://stackoverflow.com/questions/15518328/creating-js-object-with-object-createnull#answer-21079232
-		function StorageObject() {}
-		StorageObject.prototype = Object.create(null);
-
 		function _parseArray (resultSet, array) {
 			var length = array.length;
 
@@ -25,15 +20,17 @@
 		var hasOwn = {}.hasOwnProperty;
 
 		function _parseNumber (resultSet, num) {
-			resultSet[num] = true;
+			resultSet.add(num);
 		}
 
 		function _parseObject (resultSet, object) {
 			for (var k in object) {
 				if (hasOwn.call(object, k)) {
-					// set value to false instead of deleting it to avoid changing object structure
-					// https://www.smashingmagazine.com/2012/11/writing-fast-memory-efficient-javascript/#de-referencing-misconceptions
-					resultSet[k] = !!object[k];
+					if (!!object[k]) {
+						resultSet.add(k);
+					} else {
+						resultSet.delete(k);
+					}
 				}
 			}
 		}
@@ -44,7 +41,7 @@
 			var length = array.length;
 
 			for (var i = 0; i < length; ++i) {
-				resultSet[array[i]] = true;
+				resultSet.add(array[i]);
 			}
 		}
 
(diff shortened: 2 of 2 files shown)


JedWatson:
Sorry for the delayed reply @newyork-anthonyng, and thanks for the PR

I'm concerned about dropping support for IE11. I know it's a pain but there's still too much usage for that to be safe in a library that's this widely used.

However IE11 support might be good enough aside from the missing `Set.values` method. I think we could work around that by detecting `Set.prototype.values` and falling back to using [`forEach`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/forEach) to turn the result into a string.

What do you think of that?

newyork-anthonyng:
@JedWatson Good callout.

I updated and removed the `Set.prototype.values` and `Array.from` calls because of the lack of `IE11` support. Let me know if there's any other changes.

tests/dedupe.js
```diff
@@ -31,7 +31,7 @@ describe('dedupe', function () {
 	});
 
 	it('joins arrays of class names and ignore falsy values', function () {
-		assert.equal(dedupe('a', 0, null, undefined, true, 1, 'b'), '1 a b');
```
dcousens: I don't know if this is actually deterministic

tom-sherman: When you iterate through a Set you receive the values in insertion order, so this should be deterministic.

tom-sherman:
Are we able to merge this once the conflict has been resolved?

dcousens:
I will mark this as a breaking change, but I think this will tie in nicely with the other upcoming breaking changes

dcousens (approved):
LGTM

edit: needs merge

dcousens (approved):
LGTM

dcousens:
Running the benchmarks in browser and node

Node
```
* local/dedupe#strings x 1,823,058 ops/sec ±0.61% (95 runs sampled)
*   npm/dedupe#strings x 2,021,462 ops/sec ±0.96% (94 runs sampled)

* local/dedupe#object x 3,403,641 ops/sec ±0.50% (95 runs sampled)
*   npm/dedupe#object x 3,255,057 ops/sec ±2.78% (82 runs sampled)

* local/dedupe#strings, object x 1,912,406 ops/sec ±0.85% (91 runs sampled)
*   npm/dedupe#strings, object x 1,770,558 ops/sec ±1.53% (84 runs sampled)

* local/dedupe#mix x 1,318,324 ops/sec ±1.07% (93 runs sampled)
*   npm/dedupe#mix x 625,322 ops/sec ±2.56% (81 runs sampled)

* local/dedupe#arrays x 869,689 ops/sec ±0.72% (93 runs sampled)
*   npm/dedupe#arrays x 854,968 ops/sec ±1.79% (85 runs sampled)
```

For node, I can see some benefit for `local/dedupe#mix`, but everything else is within a margin of error.

`Mozilla/5.0 (X11; Linux x86_64; rv:105.0) Gecko/20100101 Firefox/105.0`
```
* local/dedupe#strings x 1,136,879 ops/sec ±0.53% (65 runs sampled)
* npm/dedupe#strings x 2,187,764 ops/sec ±0.50% (63 runs sampled)

* local/dedupe#object x 1,418,991 ops/sec ±0.97% (62 runs sampled)
* npm/dedupe#object x 3,763,277 ops/sec ±0.55% (63 runs sampled)

* local/dedupe#strings, object x 1,010,576 ops/sec ±1.02% (61 runs sampled)
* npm/dedupe#strings, object x 1,836,281 ops/sec ±1.31% (62 runs sampled)

* local/dedupe#mix x 783,200 ops/sec ±1.48% (62 runs sampled)
* npm/dedupe#mix x 604,930 ops/sec ±28.67% (41 runs sampled)
* 
* local/dedupe#arrays x 488,734 ops/sec ±6.45% (60 runs sampled)
* npm/dedupe#arrays x 772,377 ops/sec ±0.54% (64 runs sampled)
```

For my browser, I am seeing significant regressions in the ops/sec for each category _except_ `local/dedupe#mix`.
I don't think we will accept this change as written; not without a significant positive impact.

PR #179, opened 2018-10-24
exported concealed types in TypeScript declaration file

jacamera:
Just exporting the additional types that the classnames function takes as arguments. Especially useful when you want to create a reusable component that accepts a `ClassValue` as a prop.


JedWatson:
Thanks @jacamera

PR #180, opened 2018-10-24
Problems with typescript and classNames.default

zhuangzhuang:
like https://github.com/axios/axios/blob/master/lib/axios.js#L53
when use typescrit with requirejs and webpack only one can work.

### DIFF
--- a/index.js
+++ b/index.js
@@ -38,8 +38,8 @@
 		return classes.join(' ');
 	}
 
+	classNames.default = classNames;
 	if (typeof module !== 'undefined' && module.exports) {
-		classNames.default = classNames;
 		module.exports = classNames;
 	} else if (typeof define === 'function' && typeof define.amd === 'object' && define.amd) {
 		// register as 'classnames', consistent with npm package name


dcousens:
Can anyone else independently verify this?

jonkoops:
I believe this would be resolved by switching to ES modules in #243

christianheyn:
> Can anyone else independently verify this?

Yes. I have the same issue in next.js typescript project.

```sh
Could not find a declaration file for module 'classnames'. '[...]/node_modules/classnames/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/classnames` if it exists or add a new declaration (.d.ts) file containing `declare module 'classnames';`

2 import classnames from 'classnames'

```
Bildschirmfoto 2021-11-10 um 11 27 43

jonkoops:
@christianheyn could you confirm this is still an issue when you use the `next` branch?

christianheyn:
> @christianheyn could you confirm this is still an issue when you use the `next` branch?

Yes

`npm i JedWatson/classnames#next -S`

solves the issue for me.

christianheyn:
> What version of Node are you running @christianheyn?

v15.12.0

jonkoops:
I am unable to reproduce this issue using the `next` version, here are the steps I followed:

```bash
nvm install 15.12.0
npx create-next-app nextjs-test --use-npm --typescript
cd nextjs-test
npm install JedWatson/classnames#next
npm run dev
```

I then replaced the contents of `pages/index.tsx` with:

```tsx
import classNames from 'classnames'
import type { NextPage } from 'next'

const Home: NextPage = () => (<div className={classNames('foo', 'bar')}></div>)

export default Home
```

Which compiles without issue and renders the following to the page as expected:

```html
<div class="foo bar"></div>
```

@christianheyn is there some way you could share part of your project in a way that reproduces this issue?

jonkoops:
@dcousens @JedWatson I think we can consider this resolved in the `next` version, should we close the issue?

Issue #182, opened 2018-11-01
`undefined` should be returned when no classes are needed


jakubsadura:
It's quite common that we do not have any class name to be added to DOM, in such case returning `undefined` would allow having cleaner DOM, without empty `class` attributes.

Example with React's jsx:
```js
const isFoo = false;
const isBar = false;
const classes = classNames({
  'foo': isFoo,
  'bar': isBar,
});

return (<i className={classes}/>);
```

results in:
`<i class=""/>`

while it should be:
`<i/>`

@see https://github.com/facebook/react/issues/1448

DiscoNova:
I may be commenting the obvious on this old PR, but this simply will not work no various cases where classnames is still supposed to do its thing while not introducing unexpected classes to elements.

For example ... ```document.documentElement.classList.add(undefined)``` leads to, you may have guessed it ... ```<html class="undefined">``` - which quite likely isn't what the author of the code intended.

dcousens:
See #183

PR #183, opened 2018-11-01
return undefined for empty class list

jakubsadura:
Fixes https://github.com/JedWatson/classnames/issues/182

### DIFF
--- a/tests/index.js
+++ b/tests/index.js
@@ -27,8 +27,8 @@ describe('classNames', function () {
 		assert.equal(classNames('', 'b', {}, ''), 'b');
 	});
 
-	it('returns an empty string for an empty configuration', function () {
-		assert.equal(classNames({}), '');
+	it('returns undefined for an empty configuration', function () {
+		assert.equal(classNames({}), undefined);
 	});
 
 	it('supports an array of class names', function () {
--- a/index.js
+++ b/index.js
@@ -35,7 +35,10 @@
 			}
 		}
 
-		return classes.join(' ');
+		var allClasses = classes.join(' ').trim();
+		var result = allClasses.length === 0 ? undefined : allClasses;
+
+		return result;
 	}
 
 	if (typeof module !== 'undefined' && module.exports) {


dcousens:
This is a breaking change unfortunately.

If anyone assumed the returned type was `''`,  then they might be performing string concatenation based on that assumption (I know I have).

The end result is that `'undefined'` may appear.

jakubsadura:
You're right. Can wyou ship it with major version bump? I can update the docs if needed

dcousens:
I think this change is scary enough,  that I'd want confirmation from @JedWatson and other users of the package to weigh in before anything was changed or released.

JedWatson:
So in terms of avoiding unexpected behaviour, I don't think we can ship this. It's not so much the hassle of a major version, as the possible usage where `undefined` gets coerced into a string.

Imagine the following case where the usage of classNames is abstracted away in a library:

```js
// in a library
function getClassnameForButton(opts) {
  // this could return undefined
  return classnames(opts.specialClass);
}

// in user-land
const MyButton = (props) => <button class={`${getClassnameForButton(props)} my-button`} />
```

In this case the button would have a class of `undefined my-button`

That's pretty contrived but I'm sure there are more valid scenarios where the usage of `classnames` is not necessarily known to the component, and while I can totally appreciate not wanting empty class properties on components (especially for SSR) I think the priority for this library is safety and predictability.

You can always wrap it, if concatenating strings onto the result is something you'd never do in your components:

```
const cx = args => classnames(...args) || undefined;
const MyThing = ({ customClass }) => <span class={cx(customClass)} />;
```

JedWatson:
Also worth mentioning that we're talking about React here but this library is framework-agnostic, and I'm not sure we can guarantee that this isn't used in an environment where the result could be coerced into a string anyway.

vladshcherbin:
@JedWatson maybe it's possible to export a separate function which return `undefined`, `cxu` or any other name so we don't need wrappers in our projects 🤔

jakubsadura:
Good point @JedWatson
What do you think of @vladshcherbin idea?

kylemh:
@JedWatson I don't think your example is contrived, but one that's possible. More importantly, I'd say that shouldn't be a concern in terms of making an improvement to the library. That same user should be told that they should be using `classNames` in the render as well since the function could potentially return undefined.

You say that the library's priority should be safety and predictability, but I found the issue this PR attempts to solve to be unexpected. While a massive micro-optimization, you'll also be helping the web to ship smaller bundles collectively with this change.

Major version bumps are exactly for this purpose, and people can choose not to upgrade if they'd like to maintain old behavior. I just wanted to add another voice to the discussion in hopes that you'd change your mind. No worries if not.

tests/index.js
```diff
@@ -27,8 +27,8 @@ describe('classNames', function () {
 		assert.equal(classNames('', 'b', {}, ''), 'b');
 	});
 
-	it('returns an empty string for an empty configuration', function () {
-		assert.equal(classNames({}), '');
+	it('returns undefined for an empty configuration', function () {
```
lokeshjain2008: you should check if undefined will become string in the template

lokeshjain2008: empty string is better and predictable for UI developers.

kylemh: ... the whole point of this PR is to make it undefined. Read the above posts to understand why.

soluml:
The npm package has an alternate dedupe and bind version. I understand this PR as being problematic for projects that expect empty string so, thoughts on another alternate version? All it'd need to do is bring in the base class and and add the empty string check to return undefined.

I use classnames in many of my projects and in most of those I end up writing a wrapper util to do the above.  Would love to get it OOTB in a single npm install! :)

dcousens:
As is, I think this PR should close for the aforementioned _breaking change_ reasons,  but as hinted by @soluml maybe the path forward is another variant.
If you feel strongly about the feature,  I'm open to that PR,  but,  a reasonable warning is that feature development on this package is generally quite conservative.

PR #189, opened 2019-01-30
Upgrade outdated dependencies

grushetsky:
It turns out that some of dependencies contain vulnerabilities (even a critical one):

image

Changes:

- Upgrade dependencies to latest versions.
- Add lock files.
- Use `map` instead of `pluck` in benchmark project. [`pluck` was removed](https://github.com/lodash/lodash/wiki/Changelog#jan-12-2016--diff--docs) in later versions of Lodash.


JedWatson:
Thanks @grushetsky

PR #192, opened 2019-04-13
Add missing quotes in README.md example

SmolinPavel:
## Add missing quotes in README.md example

This `btn` should be a string:

image


newyork-anthonyng:
The quotes are needed for `btn-pressed` and `btn-over` because of the hyphen in its name. `btn` itself doesn't need it.

SmolinPavel:
@newyork-anthonyng oh yeah, my bad.

PR #194, opened 2019-06-07
ES Modules, but as a non-breaking change

developit:
My hope is that this represents a baby-step towards #150 that might be easier to merge, since it's not a breaking change. All of the files remain in-place, but the main module (`index.js`) also has an ES Modules version.

The effect this has is more visible when bundled via Webpack or Rollup, but here's the raw size improvement just looking at minified versions of master VS this PR:

| Current Master | After this PR | Difference
|--|--|--
| 339b | 255b | _-84b / 25%_

I know this seems like a small value! But for a module downloaded 3M times per month, that's quite a few bytes over the wire. It also helps ensure `classnames` gets concatenated effectively by Webpack and Rollup, which makes the module more likely to be moved into shared code chunks that avoid duplication.

/cc @JedWatson @Andarist


Andarist:
@developit actually this might unfortunately break many people - https://github.com/webpack/webpack/issues/7973 .

FDiskas:
This must be merged.

Andarist:
@FDiskas there is nothing that "must" be done in open source.

FDiskas:
@Andarist this is likely  unmaintained project. So we can mess around :) as we like

dcousens:
Merged https://github.com/JedWatson/classnames/pull/243

PR #196, opened 2019-07-02
Types definition improvements

xobotyi:
Main purpose is to define both, named and default export which package does;
`.d.ts` files size reduced making direct exports where it is possible;


xobotyi:
@JedWatson @dcousens will it be merged or shall i demove the PR?

burtek:
It would also be nice to finally include the types in the npm package

PR #197, opened 2019-07-02
Vast performance improvements;

xobotyi:
Reworked all implementations (regular, dedupe, bind).
No breaking changes.

Benchmarks:
```bash
> node ./benchmarks/run

* local#strings x 15,466,941 ops/sec ±0.36% (92 runs sampled)
*   npm#strings x 3,646,035 ops/sec ±1.69% (88 runs sampled)
* local/dedupe#strings x 2,648,941 ops/sec ±0.19% (93 runs sampled)
*   npm/dedupe#strings x 1,470,214 ops/sec ±1.71% (94 runs sampled)

> Fastest is local#strings

* local#object x 17,545,632 ops/sec ±0.42% (93 runs sampled)
*   npm#object x 3,899,880 ops/sec ±1.32% (93 runs sampled)
* local/dedupe#object x 7,118,462 ops/sec ±0.26% (91 runs sampled)
*   npm/dedupe#object x 2,632,530 ops/sec ±0.16% (94 runs sampled)

> Fastest is local#object

* local#strings, object x 11,379,569 ops/sec ±0.44% (95 runs sampled)
*   npm#strings, object x 3,259,972 ops/sec ±0.40% (94 runs sampled)
* local/dedupe#strings, object x 2,525,953 ops/sec ±1.49% (93 runs sampled)
*   npm/dedupe#strings, object x 1,548,802 ops/sec ±0.14% (92 runs sampled)

> Fastest is local#strings, object

* local#mix x 7,226,283 ops/sec ±0.20% (95 runs sampled)
*   npm#mix x 2,521,931 ops/sec ±1.39% (93 runs sampled)
* local/dedupe#mix x 907,851 ops/sec ±0.39% (93 runs sampled)
*   npm/dedupe#mix x 667,852 ops/sec ±0.84% (95 runs sampled)

> Fastest is local#mix

* local#arrays x 3,571,391 ops/sec ±0.17% (94 runs sampled)
*   npm#arrays x 919,980 ops/sec ±0.22% (91 runs sampled)
* local/dedupe#arrays x 1,028,421 ops/sec ±1.41% (95 runs sampled)
*   npm/dedupe#arrays x 743,585 ops/sec ±0.17% (94 runs sampled)

> Fastest is local#arrays
```

### DIFF
--- a/dedupe.js
+++ b/dedupe.js
@@ -51,7 +51,7 @@
 	function classNames(){
 		var classSet = new StorageObject();
 
-		var str = "", n,
+		var str = '', n,
 			len = arguments.length;
 
 		for (n = 0; n < len; n++) {
@@ -60,7 +60,7 @@
 
 		for (n in classSet) {
 			if (classSet[n]) {
-				str && (str += " "), (str += n);
+				str && (str += ' '), (str += n);
 			}
 		}
 
--- a/dedupe.js
+++ b/dedupe.js
@@ -1,100 +1,71 @@
 /*!
-  Copyright (c) 2018 Jed Watson.
-  Licensed under the MIT License (MIT), see
-  http://jedwatson.github.io/classnames
-*/
+ Copyright (c) 2018 Jed Watson.
+ Licensed under the MIT License (MIT), see
+ http://jedwatson.github.io/classnames
+ */
 /* global define */
 
 (function () {
 	'use strict';
 
-	var classNames = (function () {
-		// don't inherit from Object so we can skip hasOwnProperty check later
-		// http://stackoverflow.com/questions/15518328/creating-js-object-with-object-createnull#answer-21079232
-		function StorageObject() {}
-		StorageObject.prototype = Object.create(null);
+	var hasOwn = Object.prototype.hasOwnProperty;
+	var isArray = Array.isArray;
+	var SPACE = /\s+/;
+	
+	// don't inherit from Object so we can skip hasOwnProperty check later
+	// http://stackoverflow.com/questions/15518328/creating-js-object-with-object-createnull#answer-21079232
+	function StorageObject() {}
+	StorageObject.prototype = Object.create(null);
 
-		function _parseArray (resultSet, array) {
-			var length = array.length;
+	function _parse (resultSet, arg) {
+		if (!arg) return;
+		
+		var i, length;
+		
+		if (typeof arg === 'string') { // 'foo bar'
+			var array = arg.split(SPACE);
+			length = array.length;
 
-			for (var i = 0; i < length; ++i) {
-				_parse(resultSet, array[i]);
-			}
-		}
-
-		var hasOwn = {}.hasOwnProperty;
-
-		function _parseNumber (resultSet, num) {
-			resultSet[num] = true;
-		}
-
-		function _parseObject (resultSet, object) {
-			for (var k in object) {
-				if (hasOwn.call(object, k)) {
-					// set value to false instead of deleting it to avoid changing object structure
-					// https://www.smashingmagazine.com/2012/11/writing-fast-memory-efficient-javascript/#de-referencing-misconceptions
-					resultSet[k] = !!object[k];
-				}
-			}
-		}
-
-		var SPACE = /\s+/;
-		function _parseString (resultSet, str) {
-			var array = str.split(SPACE);
-			var length = array.length;
-
-			for (var i = 0; i < length; ++i) {
+			for (i = 0; i < length; ++i) {
 				resultSet[array[i]] = true;
 			}
-		}
+		} else if (isArray(arg)) { // ['foo', 'bar', ...]
+			length = arg.length;
 
-		function _parse (resultSet, arg) {
-			if (!arg) return;
-			var argType = typeof arg;
-
-			// 'foo bar'
-			if (argType === 'string') {
-				_parseString(resultSet, arg);
-
-			// ['foo', 'bar', ...]
-			} else if (Array.isArray(arg)) {
-				_parseArray(resultSet, arg);
-
-			// { 'foo': true, ... }
-			} else if (argType === 'object') {
-				_parseObject(resultSet, arg);
-
-			// '130'
-			} else if (argType === 'number') {
-				_parseNumber(resultSet, arg);
+			for (i = 0; i < length; ++i) {
+				_parse(resultSet, arg[i]);
 			}
-		}
-
-		function _classNames () {
-			// don't leak arguments
-			// https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#32-leaking-arguments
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -8,35 +8,44 @@
...
+	function reduceArray (args) {
+		var len = args.length;
+		if (!len)
+			return "";
+		var str = "", item, type, i, n;
+		for (i = 0; i < len; i++) {
+			if (!(item = args[i]))
+				continue;
+			type = typeof item;
+			if (type === "string" || type === "number") {
```
xobotyi: The only thing which i dont get - why do we have to pass the numbers here?
Due to [W3C's TR](http://www.w3.org/TR/CSS21/grammar.html#scanner) - they're NOT valid class names.

JedWatson:
Those are some big improvements to the benchmarks @xobotyi, awesome!

I'm reviewing this now and can't see any issues with merging your changes.

Out of curiosity, would you mind annotating the PR (adding your own review is a good way to do this) with some explanations of why the changes you made improved the performance?

Thanks!

index.js
```diff
@@ -8,35 +8,44 @@
 (function () {
 	'use strict';
 
-	var hasOwn = {}.hasOwnProperty;
-
-	function classNames () {
-		var classes = [];
```
xobotyi: The main idea here is to go away from arrays.  
Due to the iterative process we can replace array accumulator with simple string.  
It gives us two benefits:
- String concatenation is faster than array push.
- Because we are generating the result on the fly - there is no need of `classes.join(' ')` which in fact caused second array iteration (first was over the function arguments).

Secondary: i've moved variables declaration out of the loops (it gives pretty big impact).
Plus, IE's optimizer works better if to make `typeof` separately without storing it's result into a variable.

dedupe.js
```diff
@@ -1,100 +1,71 @@
...
 
-			// '130'
-			} else if (argType === 'number') {
-				_parseNumber(resultSet, arg);
-			}
+		for (n = 0; n < len; n++) {
+			_parse(classSet, arguments[n]);
 		}
 
-		function _classNames () {
```
xobotyi: @JedWatson almost the same optimisation reason here (but the real showstopper here is strings split, with which we can do nothing).
But here we had even worse situation in terms of iteration:
- Lines 76-80 are literally useless. It creates new, real array of `arguments` elemnts and then araing iterates over newborn array at lines 87-91; It would've made sence if you passes that array further to other functions but not here.
- And then, again, simply aviding the array accumulator while building the final string;
- Futrthermore i've simply merged parsers into single function, that gives us economy on function calls;

xobotyi:
Done!
Glad i did help😇 (actually first time contributing to such a big, in terms of users, package🤪)

index.js
newyork-anthonyng: This handles use cases like this:

```js
classNames('foo', 23); // => 'foo 23'
```

newyork-anthonyng:
The original implementation did something similar. 
This [PR](https://github.com/JedWatson/classnames/pull/50#issuecomment-144703425) changed it to the current implementation with arrays. It seemed like _readability_ was the main reason of switching over.

index.js
xobotyi: I know what it does, 23 is invalid classname (as any number leaded string will be), that's I wonder why this lib passes it to result.

xobotyi:
@newyork-anthonyng looks kinda weird to me, to speculate readability of code you'll never read with obvious performance  tradeoff :japanese_goblin:

dcousens:
Will inspect this soon :+1:

index.js
dcousens: Could you quantify the performance differences of everything _other_ than the String concatenation change?
We have had nothing but to-and-fro changes between the two methods on this project.

The results appear to vary on every build of V8.

xobotyi: @dcousens it impacts not only the V8 based browsers, FF has performance improvements too.

index.js
```diff
@@ -8,34 +8,44 @@
...
+
+		if (!len) return '';
+
+		var str = '', item, i, n;
+		for (i = 0; i < len; i++) {
+			if (!(item = arguments[i]))	continue;
+
+			if (typeof item === 'string' || typeof item === 'number') {
+				str && (str += ' ');
+				(str += item);
```
krawaller: Since we're chasing milliseconds, won't it be faster to just assign once?

```
str = (str ? ' ' : '') + n;
```

krawaller:
It very likely won't make a measurable difference, but I'm curious if string constants would be an improvement.

```javascript
var EMPTY = ''
var SPACE = ' ' // (except regex already called this)
var OBJECT = 'object'
var FUNCTION = 'function'
// ...etc
```

It might well be that variables perform worse than literals (when not using `const`) since they are mutable and literals aren't.

xobotyi:
@krawaller there are no perf improvements if to store them in variables, even more - it will slow down the code due to disabling the typecheck engine optimisation;
```js
typeof a === 'string'
``` 
is always faster than 
```js
var STR = 'string';
typeof a === STR;
```
or even
```js
var type = typeof a;
type === 'string'
```

index.js
xobotyi: Practically, for v8 my aproach is faster, for other browsers they're almost equal;
I suppose it is caused by prediction branching.
In your way you have 2 concats all the time, in mine - 1 guaranteed and one optional;

krawaller:
Ah, roger that, @xobotyi!

index.js
krawaller: Yes, but my gut feeling was that the extra concat done on item 0 is worth avoiding the extra variable mutations for items 1+.

xobotyi: @krawaller benchmarks result:
mine:  `local#strings x 14,682,684 ops/sec ±1.21% (89 runs sampled)`
yours: `local#strings x 14,028,615 ops/sec ±1.01% (89 runs sampled)`

It is obviusly a very small difference but if it can be done - why not =)

krawaller: Yeah, I guess we've left the real world far behind. :)

Still, very interesting! I really thought variable mutation would be more expensive, but yours win out. Predictive branching for the win, I guess!

xobotyi: @krawaller tbh first time i've seen the results i've been wondered too =)
First iteration code been written as yours %)

robertmaier:
Are there any plans to merge this?

zombieJ:
@JedWatson , ping again. Pls help to review and merge. Thank.

ref: https://github.com/ant-design/ant-design/pull/21440

resynth1943:
cc: @JedWatson any updates on this, please?

pegak:
Any updates? Is there anything blocking the perf improvement?

xobotyi:
@pegak nothing, imo.
I can offer you a drop-in replacement https://github.com/xobotyi/cnbuilder

PR #218, opened 2020-07-20
Refactor `classNames` to use an internal closure

Lucretiel:
Refactor `classNames` to use an internal closure, so that all classes are appended to a single `classes` array. This reduces the creation of intermediary strings.


Lucretiel:
Looks like #197 is the preferable refactor.

PR #229, opened 2021-04-01
Minor cleanup

dcousens:
Firstly, this PR changes the first branch when unpacking object's, back to the primary behavior, with the `toString` behaviour as an addendum,  and fixes some of the mixed indentation that was merged previously.

Secondly, after updating the developer dependencies, I removed the `open-cli` test target and dependency,  as it attempts to `open` in a compatible program, but squashes any errors.
I don't think opening the default browser is useful when testing compared to maybe providing instructions in the README.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -8,15 +8,10 @@
 //   Michal Adamczyk <https://github.com/mradamczyk>
 //   Marvin Hagemeister <https://github.com/marvinhagemeister>
 
-export type ClassValue = string | number | boolean | undefined | null | ClassDictionary | ClassArray;
+export type Value = string | number | boolean | undefined | null;
+export type Mapping = { [key: string]: Value };
+export type Argument = Value | Mapping | Argument[];
 
-export interface ClassDictionary
-{
-	[id: string]: boolean | undefined | null;
-}
-
-export interface ClassArray extends Array<ClassValue> {}
-
-export function classNames(...args: ClassArray): string;
+export function classNames(...args: Argument[]): string;
 
 export default classNames;
--- a/index.d.ts
+++ b/index.d.ts
@@ -8,7 +8,7 @@
 //   Michal Adamczyk <https://github.com/mradamczyk>
 //   Marvin Hagemeister <https://github.com/marvinhagemeister>
 
-export type ClassValue = string | number | ClassDictionary | ClassArray | undefined | null | boolean;
+export type ClassValue = string | number | boolean | undefined | null | ClassDictionary | ClassArray;
 
 export interface ClassDictionary
 {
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -26,9 +25,8 @@
     "utility"
   ],
   "devDependencies": {
-    "benchmark": "2.1.4",
-    "browserify": "16.2.3",
-    "mocha": "5.2.0",
-    "opn-cli": "4.0.0"
+    "benchmark": "^2.1.4",
+    "browserify": "^16.2.3",
+    "mocha": "^8.3.2"
```
dcousens, line 30: Updated `mocha` too, the listed version actually threw errors on startup for me.

dcousens:
@JedWatson tests are passing, but the Travis CI for this repository appears to have disappeared.

index.d.ts
```diff
@@ -8,15 +8,10 @@
...
+export type Argument = Value | Mapping | Argument[];
 
-export interface ClassDictionary
-{
-	[id: string]: boolean | undefined | null;
-}
-
-export interface ClassArray extends Array<ClassValue> {}
-
-export function classNames(...args: ClassArray): string;
```
dcousens, line 20: Removed the `Class` prefix (redundant when you are importing).

The mapping of `string` to truthy values was different between between arrays and objects,  which is not what our tests dictate.   Changed that.

index.d.ts
```diff
@@ -8,15 +8,10 @@
 //   Michal Adamczyk <https://github.com/mradamczyk>
 //   Marvin Hagemeister <https://github.com/marvinhagemeister>
 
-export type ClassValue = string | number | ClassDictionary | ClassArray | undefined | null | boolean;
+export type Value = string | number | boolean | undefined | null;
+export type Mapping = { [key: string]: Value };
+export type Argument = Value | Mapping | Argument[];
```
dcousens, line 13: I don't think we necessarily needed to `export` these... but, if it's useful,  why not.

index.d.ts
```diff
@@ -8,15 +8,10 @@
 //   Michal Adamczyk <https://github.com/mradamczyk>
 //   Marvin Hagemeister <https://github.com/marvinhagemeister>
 
-export type ClassValue = string | number | ClassDictionary | ClassArray | undefined | null | boolean;
+export type Value = string | number | boolean | undefined | null;
```
dcousens, line 11: I don't know if `Value` is a nice name to be honest.

These types aren't exactly capturing the subtlety of `string` arguments, compared to value truthiness.

JedWatson (approved):

dcousens:
For posterity, I benchmarked the `2.3.0` release against `2.2.6` [in Firefox 87] and there was negligible difference.

PR #234, opened 2021-04-01
Fix value types

remcohaszing:
A mapping consistes of classnames mapped to a value checked its booliness to
determine if the classnames should be added. These values may be anything,
they’re not related to the allowed class values.


OlEzhA0:
+1

dkaraush:
+1, it partially solves my issue #235

dcousens:
Thanks @remcohaszing,  published as `classnames@2.3.1`.

Issue #240, opened 2021-04-14
Class names being returned as `[object Object]` as of v2.3.0


BrianAtIgloo:
Hi, first off - thank you for maintaining this fantastic package!

That said, we've run into an issue with v`2.3.0` where our bundled components are now outputting `[object Object]` class names because of [this recent change](https://github.com/JedWatson/classnames/pull/170) to support a custom `toString` fn.

The problem only occurs after we run our code through our build pipeline which uses Rollup/Babel to produce a CJS bundle so we can render server-side.

Component Code:
```js
  const inputStyleClass = classNames(styles['text_input__input'], {
    [styles['text_input__input--with-icon']]: !!icon,
  });
```

Transpiled code:
```js
  var inputStyleClass = classnames(
    styles$z['text_input__input'],
    (_classNames2 = {},
        _defineProperty$2(_classNames2, styles$z['text_input__input--with-icon'], !!icon),
        _classNames2,
    )
  );
```

We've locked our components to `2.2.6` to resolve the issue and suggest the following change:

```js
    if ((/\{\s*\[native code\]\s*\}/).test(arg.toString)) {
      for (var key in arg) {
        if (hasOwn.call(arg, key) && arg[key]) {
          classes.push(key);
        }
      }
    } else {
      classes.push(arg.toString());
    }
```

dcousens:
> The problem only occurs after we run our code through our build pipeline which uses Rollup/Babel to produce a CJS bundle so we can render server-side.

Is that to imply that the pipeline is mangling the resultant object's `toString`?

dcousens:
@BrianAtIgloo could you verify if `.hasOwnProperty('toString')` resolves your issue?

BrianAtIgloo:
Thank you for the quick reply + troubleshoot! I confirmed using `hasOwnProperty` works, I can open a PR later today 👍

markdalgleish:
We're experiencing this issue because we're executing our render code in a Node VM, which means we actually have multiple Object globals in memory and the `toString` equality check fails. I managed to create a minimal reproduction: https://runkit.com/embed/3svylngmhh23

dcousens:
Fixed in https://github.com/JedWatson/classnames/pull/281

PR #241, opened 2021-04-14
Bugfix - Class names being returned as [object Object]

BrianAtIgloo:
Updated to use `hasOwnProperty` over trying to compare the Object prototype which fails with some build pipelines after transpilation. See issue #240.

### DIFF
--- a/index.js
+++ b/index.js
@@ -29,7 +29,7 @@
 					}
 				}
 			} else if (argType === 'object') {
-				if (arg.toString === Object.prototype.toString) {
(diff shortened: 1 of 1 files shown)


dcousens (approved):
LGTM ~waiting on CI~

dcousens:
This may not be as straight forward, there are issues with inheritance in my testing.

dcousens:
Reverted in https://github.com/JedWatson/classnames/commit/94c25acf7007293a673acf181737a0737b4ab134

PR #231, opened 2021-04-01
add nodejs test workflow

dcousens:
This PR replaces the existing CI integrations, with GitHub workflows

PR #242, opened 2021-04-19
Add GitHub Actions and run tests for all supported Node versions

jonkoops:
Replaces Travis with GitHub Actions and runs the tests for all actively [supported versions](https://nodejs.org/en/about/releases/) of Node.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,9 @@
     "url": "https://github.com/JedWatson/classnames.git"
   },
   "types": "./index.d.ts",
+  "engines": {
+    "node": "^10 || ^12 || ^14 || ^15"
+  },
   "scripts": {
     "benchmarks": "node ./benchmarks/run",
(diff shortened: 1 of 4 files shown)


dcousens:
Sorry @jonkoops,  I had [this pull request](https://github.com/JedWatson/classnames/pull/231) in draft and ready.
Your contribution is appreciated but we had already made the changes.

package.json
```diff
@@ -10,6 +10,9 @@
     "url": "https://github.com/JedWatson/classnames.git"
   },
   "types": "./index.d.ts",
+  "engines": {
+    "node": "^10 || ^12 || ^14 || ^15"
+  },
```
dcousens, line 15: I don't know if we want to limit the `engines` in this way

jonkoops:
No prob, missed that one!

package.json
jonkoops: It doesn't prevent the package from being installed to my knowledge. NPM will simply produce a warning.

PR #244, opened 2021-04-19
Fix misconfigured 'fail-fast' option

jonkoops:
Fixes a configuration error so that GitHub Actions will run correctly.


dcousens:
Ref: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

dcousens:
Thanks @jonkoops

PR #248, opened 2021-05-05
TypeScript <3.7 support by self-reference type rewrite

coadan:


coadan (commented):
Just squash when merging to master, I realized that the type that I suggested in the issue was not entierly correct when reviewing the code after my morning coffee. The array type should be part of its own generic and the ArgumentArray in the Argument type should not have [] symbol after it.

dcousens (approved):
LGTM, waiting on CI and then I'll try this out locally!

Thanks @coadan :+1:

PR #252, opened 2021-05-18
Drop support for NodeJS versions older than 12

jonkoops:


jonkoops:
@dcousens Could you create a `next` branch we discussed before so that I can target these changes there? This change is in preparation of the changes in #243, which I will also direct to the `next` branch so we can accumulate breaking changes.

package.json
```diff
@@ -10,6 +10,9 @@
     "url": "https://github.com/JedWatson/classnames.git"
   },
   "types": "./index.d.ts",
+  "engines": {
+    "node": ">=12.22.0"
```
jonkoops, line 14: This is the minimum version of Node that has support for ES modules (see: https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V12.md#2021-03-30-version-12220-erbium-lts-richardlau)

dcousens:
Created the `next` branch :+1:

dcousens:
Does this need to be in the `next` branch though?
I think we are happy to drop Node 0.10 as it is EOL?

jonkoops:
> Does this need to be in the `next` branch though?
> I think we are happy to drop Node 0.10 as it is EOL?

Yeah, depends on what we would consider a breaking change. If we drop support for older versions of Node explicitly this is kind of a breaking change I guess?

In this case it's Node 10 by the way, not 0.10. There are probably still some using it in the wild.

jonkoops:
@dcousens I've gone ahead and changed the base to `next`, the CI is looking all :heavy_check_mark: so merge at your convienience. I'll go ahead and rebase #243 on top of this in the meantime.

package.json
dcousens, line 14:
```suggestion
    "node": "^12.22 || >= 14.17"
```

As per your suggestion in https://github.com/JedWatson/classnames/pull/243#discussion_r635640583,  we can drop that to `12.20` and `14.13` if that we want

dcousens (approved):

dcousens:
I think we could probably merge this to `master`,  but I agree, we haven't established plainly that we are following Node's LTS schedule implicitly;  and thereby this may be a breaking change,  so I'll `next` it for now.

PR #274, opened 2022-05-18
SSR Support

9001-Sols:
This adds a check for `window` to support SSR uses.

### DIFF
--- a/index.js
+++ b/index.js
@@ -53,6 +53,8 @@
 			return classNames;
 		});
 	} else {
-		window.classNames = classNames;
+		if (typeof window !== 'undefined') {
+			window.classNames = classNames;
+		}
 	}
 }());
--- a/index.js
+++ b/index.js
@@ -52,9 +52,7 @@
 		define('classnames', [], function () {
 			return classNames;
 		});
-	} else {
-		if (typeof window !== 'undefined') {
-			window.classNames = classNames;
-		}
+	} else if (typeof window !== 'undefined') {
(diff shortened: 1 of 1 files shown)


dcousens:
@9001-Sols why aren't you using `node_modules`?

9001-Sols:
> @9001-Sols why aren't you using `node_modules`?

Sorry I don't follow. Currently we import it via:

`package.json`:
```
  "dependencies": {
    "classnames": "^2.2.6",
    ...
```

```
import classNames from 'classnames';
```

Just the existence of `window` without the `typeof` check causes SSR frameworks to breakdown. It's currently happening with Shopify `Hydrogen` (built on `Vite`). I haven't tested other frameworks like `Next`.

dcousens:
@9001-Sols the change itself is non-controversial, but, I don't understand what you mean by
> causes SSR frameworks to breakdown

The `window.classNames =` branch should never come to pass - if `module.exports` is defined - in a server environment.

gdostie:
In a ES module context, `module.exports` does not exist https://nodejs.org/api/esm.html#no-require-exports-or-moduleexports

dcousens:
We do have an ESM branch [`next`](https://github.com/JedWatson/classnames/tree/next), but it's low priority typically

gdostie:
I think the demand is definitely there, but currently it might be hard to see. When users see ESM support is missing from `classnames`, my guess is that people will just switch to https://www.npmjs.com/package/clsx.

PR #276, opened 2022-07-26
Fix TypeScript types

remcohaszing:
The package uses commonjs, but it’s typed as if it contains faux ESM. This was problematic for people using `require` or the new `node16` module resolution.

### DIFF
--- a/bind.d.ts
+++ b/bind.d.ts
@@ -1,5 +1,9 @@
-export * from '.';
-import { ArgumentArray } from '.';
+import { ArgumentArray } from './index.js';
 
-export type Binding = Record<string, string>;
-export default function classNames(this: Binding | undefined, ...args: ArgumentArray): string;
+declare namespace classNames {
+	type Binding = Record<string, string>;
+}
+
+declare function classNames(this: classNames.Binding | undefined, ...args: ArgumentArray): string;
+
+export = classNames;
--- a/dedupe.d.ts
+++ b/dedupe.d.ts
@@ -1,2 +1,2 @@
-export * from '.';
-export { default } from '.';
+import classNames = require('./index.js');
+export = classNames;
--- a/index.d.ts
+++ b/index.d.ts
@@ -7,14 +7,17 @@
 //   Sean Kelley <https://github.com/seansfkelley>
 //   Michal Adamczyk <https://github.com/mradamczyk>
 //   Marvin Hagemeister <https://github.com/marvinhagemeister>
-// TypeScript Version: 3.0
 
-export type Value = string | number | boolean | undefined | null;
-export type Mapping = Record<string, unknown>;
-export interface ArgumentArray extends Array<Argument> {}
-export type Argument = Value | Mapping | ArgumentArray;
+declare namespace classNames {
+  type Value = string | number | boolean | undefined | null;
+  type Mapping = Record<string, unknown>;
+  interface ArgumentArray extends Array<Argument> {}
+  type Argument = Value | Mapping | ArgumentArray;
+}
 
 /**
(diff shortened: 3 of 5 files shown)


dzearing:
Please @JedWatson could you help merge/publish this? This is blocking cases where TypeScript "nodenext" resolution is being used. See this issue for details: https://github.com/microsoft/TypeScript/issues/49189

adidahiya:
Hi, FYI this is a breaking change for TypeScript consumers which is unexpected in a patch version of the classnames package.

Before, we could `import classNames from "classnames"` without enabling `esModuleInterop` in TS compilation options. Now, that import syntax is no longer valid unless we enable `esModuleInterop: true`.

I understand that this change was made to support a newer TS resolution option, but changes to typings like this should be done carefully to avoid breaking existing consumers.

May I ask why you are using `"moduleResoultion": "nodenext"`? Are you running classnames in Node.js / isomorphic rendering? IMO that's a less common use case than using classnames in a frontend-only bundle, where node resolution is _not_ used/necessary. So this "fix" may have broken usage for a larger group of consumers than it actually fixed anything for.

remcohaszing:
I see how this affects you and why you consider this breaking. However, I believe this was a bug fix. Any change of behaviour can break code, including bug fixes. You were previously relying on this bug.

According to the readme, correct usage of this package is:

```js
// CJS
const classNames = require('classnames')
```

```js
// ESM
import classNames from 'classnames'
```

However, before this PR, according to the TypeScript type definitions, correct usage was:

```js
// CJS
const classNamesExports = require('classnames')
const classNames = classNamesModule.default
```

```js
// ESM
import classNamesExports from 'classnames'
const classNames = classNamesModule.default
```

The latter technically works too, because `classnames` defines `module.exports.default`. The `default` property was added in 915186aaa510e0c6be6c2cf20788aac474790e27, but no context is given. My guess is this was to please people who didn’t configure TypeScript properly.

The latter didn’t work because it was typed correctly. It worked by accident. The situation you are describing is usage of this package inside a module bundler that allows you to import `module.exports` as default. This means you don’t rely on the property `module.exports.default`. To support such situations, TypeScript has the option [`allowSyntheticDefaultImports`](https://www.typescriptlang.org/tsconfig#allowSyntheticDefaultImports). Note that you don’t need [`esModuleInterop`](https://www.typescriptlang.org/tsconfig#esModuleInterop), because you’re not compiling to CJS.

> May I ask why you are using `"moduleResolution": "nodenext"`? Are you running classnames in Node.js / isomorphic rendering?

The JavaScript ecosystem is slowly moving towards ESM. In order to be able to use newer packages or updates, one has to keep up. `"moduleResolution": "node16"` adds support for various cases that weren’t previously supported. It it also more strict, and highlights bugs like the one fixed by this PR.

> IMO that's a less common use case than using classnames in a frontend-only bundle, where node resolution is not used/necessary.

This isn’t a matter of opinion, but of statistics that aren’t presented. I’m inclined to believe this is correct though. What matters is correctness though, not usage statistics.

> So this "fix" may have broken usage for a larger group of consumers than it actually fixed anything for.

This didn’t break usage. It highlights a problem in those consumers’ configuration. You would’ve run into the same problem eventually if you added another properly typed dependency.

For more information, see https://github.com/DefinitelyTyped/DefinitelyTyped#a-package-uses-export--but-i-prefer-to-use-default-imports-can-i-change-export--to-export-default

remcohaszing:
I stand by my previous comment, but I learnt that it **is** possible to also support the `.default` property, allowing default imports to work without `allowSyntheticDefaultImports`. Also support for UMD was missing in this PR. Both are added in #301.

PR #285, opened 2022-09-16
added dev-dep nyc to create coverage reports

achesin:
Hi there, I'm interested in seeing coverage reports for this projects so I installed `nyc` to work with the existing tools to create them. I noticed that there was an already ignored `coverage` folder, so I directed any output from `nyc` to enter that folder. 

Would you be open to running the test command with coverage as part of the build? 

Thanks!

### DIFF
--- a/.nycrc.json
+++ b/.nycrc.json
@@ -1,0 +1,3 @@
+{
+  "temp-dir": "./coverage"
+}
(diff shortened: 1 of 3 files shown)


jonkoops:
Thanks for the PR! I think I will be doing some refactor work on the tests soon, so this is likely going to end up in a conflicting state. I'll close this PR for now, but thanks again for taking the time to whip this up!

PR #301, opened 2023-03-06
Support default export and UMD

remcohaszing:
It’s possible to support both a CJS export and a `.default` property by defining the export as an interface that has a call expression and a default property. This makes it possible to use `import classNames from 'classnames'` or `classNames.default('')`.

Also this packages exposes a UMD module, meaning `classNames` is available as a global when using a TypeScript script, but not when using a TypeScript module. This is supported by adding `export as namespace classNames`.

Technically this also declares it’s valid to use `classNames.default` when using `classNames` as a global. It may be better to move the line `classNames.default = classNames` outside of the if-statement at runtime.

### DIFF
--- a/bind.d.ts
+++ b/bind.d.ts
@@ -4,6 +4,14 @@ declare namespace classNames {
 	type Binding = Record<string, string>;
 }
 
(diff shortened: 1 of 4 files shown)


dcousens:
Thanks @remcohaszing!  :blue_heart:

PR #307, opened 2023-04-27
(Typescript) Allow `ReadonlyArray`s as valid classNames parameters

BenGearset:
Simple reproduction:
```typescript
const array = ['a', 'b', 'c'] as const;
classNames(array);
```
Error:
```
TS2345: Argument of type '[readonly ["a", "b", "c"]]' is not assignable to parameter of type 'ArgumentArray'.   Type 'readonly ["a", "b", "c"]' is not assignable to type 'Argument'.     Type 'readonly ["a", "b", "c"]' is not assignable to type 'Mapping'.       Index signature for type 'string' is missing in type 'readonly ["a", "b", "c"]'.
```

This PR adds a `ReadonlyArray<Argument>` to the `Argument` union type allowing `ReadonlyArray`s to be used with `classNames`

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -12,7 +12,8 @@ declare namespace classNames {
   type Value = string | number | boolean | undefined | null;
   type Mapping = Record<string, unknown>;
   interface ArgumentArray extends Array<Argument> {}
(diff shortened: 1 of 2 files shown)


dcousens (approved):
Thanks @BenGearset

PR #316, opened 2023-09-02
Fixes #313

### LINKED ISSUE #313: When no classes are selected, classNames should return null
ClassNames with no classes selected return empty string (''). For using it in React with jsx, it would be much better to return `null`.
Example:
```jsx
<p className={classNames({ first: false, second: false)>
Hello World!
</p>
```
This results in `<p class>Hello World!</p>`, while `<p>Hello World!</p>` would be correct. The second would happen if classNames returned null in such case.

BoberGame:
**Changes:**
- If the final set of classes is empty, the function returns null instead of an empty string
In this case, if we specify a class via className, the html will not display the empty class attribute.
- Adjusted the tests for this change

**All the tests pass**

**Benchmark results:**
image

Closes #313

### DIFF
--- a/index.js
+++ b/index.js
@@ -42,7 +42,12 @@
 			}
 		}
 
-		return classes.join(' ');
+		var totalClasses = classes.join(' ');
+
+		if (totalClasses.length > 0) {
+			return totalClasses;
+		}
+		return null;
 	}
 
 	if (typeof module !== 'undefined' && module.exports) {
--- a/index.js
+++ b/index.js
@@ -42,12 +42,7 @@
 			}
 		}
 
-		var totalClasses = classes.join(' ');
-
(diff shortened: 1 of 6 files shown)


index.js
```diff
@@ -42,7 +42,12 @@
 			}
 		}
 
-		return classes.join(' ');
+		var totalClasses = classes.join(' ');
+
+		if (totalClasses.length > 0) {
+			return totalClasses;
+		}
+		return null;
```
wojtekmaj, line 50:
```suggestion
		return classes.join(' ') || null;
```

jonkoops:
I am closing this PR as it has become conflicted with `main`. We will also have to have a bit more discussion on what version 3.0 should look like before we start including breaking changes.

PR #324, opened 2023-12-21
Update Node.js version matrix on CI

jonkoops:
Updates the Node.js version matrix on CI with the [actively supported versions](https://nodejs.org/en/about/previous-releases).


jonkoops:
On closer inspection, I think I will clean up the workflows a bit and make this part of a larger PR.

PR #326, opened 2023-12-21
Replace Mocha with Node.js' built-in test runner

jonkoops:
Replaces Mocha with Node.js' [built-in test runner](https://nodejs.org/api/test.html), reducing the amount of dependencies that need to be installed.


dcousens:
Awesome

PR #327, opened 2023-12-21
Convert tests to use modern JavaScript

jonkoops:
Converts the tests to use modern JavaScript syntax, including ES modules for imports.


dcousens:
LGTM

PR #329, opened 2023-12-22
Move benchmarks into dedicated NPM workspace

jonkoops:
Moves the `classnames-benchmarks` package into an NPM workspace containing all of it's own dependencies, and creates aliases for the local and last published NPM version in the package, removing the need to provide a relative path, and reducing the need for additional NPM installs.


dcousens:
LGTM

KennanChan:
It seems that this PR causes yarn to warn about "Workspaces can only be enabled in private projects.".

jonkoops:
> It seems that this PR causes yarn to warn about "Workspaces can only be enabled in private projects.".

Don't use Yarn for this project, it has been set up to work with NPM, it's also the only package manager a lockfile is checked in for.

KennanChan:
'Classnames' is actually an indirect dependency of my project. I managed to bypass this warning by setting `workspaces-experimental false` in yarnrc.

dcousens:
@jonkoops the `package.json` is published though?

jonkoops:
> the `package.json` is published though?

Correct, but workspaces should have no influence on that as far as I know. Not sure what is going on here, will do some research.

jonkoops:
@KennanChan this should now be fixed by #350 as soon as [the release](https://github.com/JedWatson/classnames/actions/runs/7357081123) for v2.5.1 goes through.

PR #330, opened 2023-12-22
Replace `dtslint` with `tsd`

jonkoops:
Replaces `dtslint` which is no longer maintained with [`dst`](https://github.com/tsdjs/tsd), as recommended by the [migration guide](https://github.com/microsoft/DefinitelyTyped-tools/tree/main/packages/dtslint#just-looking-for-expecttype-and-expecterror).


dcousens:
LGTM

PR #332, opened 2023-12-22
Replace Browserify with Rollup

jonkoops:
Replaces Browserify with Rollup, which is compatible with ESM (will become important later). Also adds a web server to serve the benchmark from a secure origin.


dcousens:
LGTM

PR #333, opened 2023-12-23
Use modern JavaScript syntax for benchmarks

jonkoops:
Changes the benchmark code so that modern JavaScript syntax is used where possible, including modules.


benchmarks/runInBrowser.js
jonkoops, line 1: I've also changed the browser version so it uses a 'start' button for the benchmark, this makes it easier to develop, as the benchmark won't block the browser.

dcousens:
LGTM

PR #334, opened 2023-12-23
Replace `benchmark` with `tinybench`

jonkoops:
Replaces `benchmark` with [`tinybench`](https://github.com/tinylibs/tinybench), which is actively maintained and much more popular. This PR also simplifies the benchmarking logic by consolidating the code for Node.js and browsers as much as possible.


dcousens:
LGTM

PR #335, opened 2023-12-23
Run benchmarks on CI

jonkoops:
Runs the benchmarks on CI. This is used as a simple smoke test to ensure the benchmarks do not break, and to prevent big regressions from getting introduced by accident.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -52,3 +52,23 @@ jobs:
 
       - name: Check type definitions
         run: npm run check-types
+
+  benchmarks:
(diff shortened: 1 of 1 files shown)


dcousens:
>  and to prevent big regressions from getting introduced by accident.

How can this prevent performance regressions?

jonkoops:
> How can this prevent performance regressions?

It naturally doesn't automatically it by itself, but it's a good indicator when reviewing a PR.

dcousens:
@jonkoops sorry, of course you can compare the diff between the two versions from the CI output, LGTM

PR #342, opened 2023-12-27
Add `exports` field to package

### LINKED ISSUE #321: Can we add exports field for this package?
Hi,  i try to use classnames in nodejs and i need to import `classnames/bind` but not `classnames/bind.js` (because of my bundler config)

can we add something like this:
```json
{
  "exports": {
    ".": {
      "default": "./index.js"
    },
    "./bind": {
      "default": "./bind.js"
    },
    "./dedupe": {
      "default": "./dedupe.js"
    }
  },
}
```

in package.json of classnames?

jonkoops:
Adds the [`exports` field](https://nodejs.org/api/packages.html#exports) to the package, which will explicitly define the exported surface of the library. Closes #321.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,14 +2,42 @@
   "name": "classnames",
   "version": "2.4.0",
   "description": "A simple utility for conditionally joining classNames together",
-  "main": "index.js",
   "author": "Jed Watson",
   "license": "MIT",
   "repository": {
     "type": "git",
     "url": "https://github.com/JedWatson/classnames.git"
   },
+  "type": "commonjs",
+  "main": "./index.js",
   "types": "./index.d.ts",
+  "exports": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -2,14 +2,42 @@
...
   "version": "2.4.0",
   "description": "A simple utility for conditionally joining classNames together",
-  "main": "index.js",
   "author": "Jed Watson",
   "license": "MIT",
   "repository": {
     "type": "git",
     "url": "https://github.com/JedWatson/classnames.git"
   },
+  "type": "commonjs",
```
jonkoops, line 11: Added this explicit `type` field as well, which will help eliminate guesswork in future Node.js versions.

dcousens:
LGTM

PR #345, opened 2023-12-28
Add GitHub provenance for publishing

dcousens:


.github/workflows/release.yml
```diff
@@ -27,3 +30,4 @@ jobs:
         run: npm publish
```
jonkoops, line 30: Looks like the [recommended way](https://docs.npmjs.com/generating-provenance-statements#publishing-packages-with-provenance-via-github-actions) to enable provenance is to pass the `--provenance` flag directly, rather than passing it as an environment variable. But both will work.

jonkoops (approved):
LGTM

.github/workflows/release.yml
dcousens: @jonkoops I noticed that, but as someone who is using `pnpm`, this is what I'm used to

jonkoops: Indeed, if it works it works :)

PR #350, opened 2023-12-29
Remove benchmarks from workspace

jonkoops:
Publishing the package with the `workpaces` field causes issues in legacy Yarn (see https://github.com/JedWatson/classnames/pull/329#issuecomment-1871740033), removing the field and moving all the benchmark dependencies into the root should fix this issue.


dcousens:
LGTM

PR #362, opened 2024-01-08
Use string concatenation to speed up dedupe

jonkoops:
Changes the dedupe variant so that the resulting class is built up using string concatenation, rather than joining an array. This is similar to the performance enhancements implemented under #336.

## Benchmarks (compared to `main`)

This yields significant performance improvements, or in the worst case equivalent performance. The following benchmarks were ran on my 2021 MacBook Pro 14-inch with an M1 Pro chip running macOS Sonoma (14.2.1):

<details>
<summary>Node.js (v21.5.0)</summary>

Benchmarking 'strings'.

| Task Name    | ops/sec   | Average Time (ns)  | Margin | Samples |
| ------------ | --------- | ------------------ | ------ | ------- |
| dedupe/local | 4,977,383 | 200.90875648794707 | ±0.55% | 2488692 |
| dedupe/main  | 4,092,089 | 244.37392188339444 | ±0.53% | 2046045 |

Benchmarking 'object'.

| Task Name    | ops/sec   | Average Time (ns)  | Margin | Samples |
| ------------ | --------- | ------------------ | ------ | ------- |
| dedupe/local | 9,563,547 | 104.56371003714477 | ±0.46% | 4781774 |
| dedupe/main  | 7,027,889 | 142.29022110478644 | ±0.56% | 3513945 |

Benchmarking 'strings, object'.

| Task Name    | ops/sec   | Average Time (ns)  | Margin | Samples |
| ------------ | --------- | ------------------ | ------ | ------- |
| dedupe/local | 5,042,013 | 198.33345643226724 | ±0.55% | 2521007 |
| dedupe/main  | 4,292,539 | 232.96234816653524 | ±0.50% | 2146270 |

Benchmarking 'mix'.

| Task Name    | ops/sec   | Average Time (ns) | Margin | Samples |
| ------------ | --------- | ----------------- | ------ | ------- |
| dedupe/local | 2,167,627 | 461.3338635593327 | ±0.71% | 1083814 |
| dedupe/main  | 1,963,135 | 509.3892965132103 | ±0.54% | 981568  |

Benchmarking 'arrays'.

| Task Name    | ops/sec   | Average Time (ns)  | Margin | Samples |
| ------------ | --------- | ------------------ | ------ | ------- |
| dedupe/local | 2,126,593 | 470.23572153389307 | ±0.92% | 1063297 |
| dedupe/main  | 1,877,663 | 532.5767602717776  | ±0.92% | 938832  |
</details>

<details>
[...]

PR #363, opened 2024-01-08
Use string concatenation to speed up dedupe

jonkoops:
Backports #362


jonkoops:
@dcousens could you update the configuration so we can [release this version](https://github.com/JedWatson/classnames/actions/runs/7453672472)?

Issue #372, opened 2024-01-31
Please use GitHub Releases


ctnkaan:
Hello I've noticed that there are no releases for newer versions, there are only tags.

It would be great if there are notes on each new version release so people can know what is new or if they need to change something with their code when there is a major update.

I would love to help with the documentation for releases if there is an opening.

Cheers!

jonkoops:
There are release notes in the [HISTORY file](https://github.com/JedWatson/classnames/blob/main/HISTORY.md)

dcousens:
The project hasn't traditionally used GitHub releases, but I concur it is nice if you use GitHub's fine grained `watch` features.

Pull requests are welcomed to add that to our https://github.com/JedWatson/classnames/blob/main/.github/workflows/release.yml workflow

jonkoops:
Agreed, we can do some work for this. If anyone would like to contribute, feel free to open a PR. Perhaps we should also rename the file to `CHANGELOG.md`, which most dependency bots (e.g. Dependabot, Renovate) support.

ctnkaan:
I would love to contribute! Will look into this today after work.

ctnkaan:
Hey @dcousens and @dcousens 

I've opened a PR about this issue. Feel free to give any improvement ideas whenever you have time 😁

jonkoops:
I'll take a look @ctnkaan, sorry for the delay, I've been rather backlogged recently.

PR #374, opened 2024-02-14
Add Create Release Job

ctnkaan:
Related to #372 

Me and @0xD3n worked on creating this pipeline.

- Changed History.md to CHANGELOG.md
- fetch-npm-version: Fetches the version from npm
- create-github-release: Creates an new release with the title of the version number and hyperlink to CHANGELOG.md

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,7 +8,7 @@ permissions:
 
 jobs:
   publish:
-    name: Publish package to NPM
+    name: Publish package to NPM and Create GitHub Release
     runs-on: ubuntu-latest
     environment: release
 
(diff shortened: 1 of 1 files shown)


jonkoops (changes requested):
Couple of notes:

1. I think we can put all the work in a single job (under `publish`)
2. Getting the version of the package should be possible locally, using `npm pkg get version | xargs` or `cat package.json | jq -r '.version'`. This also prevents issues when we're releasing patches for older versions, rather than just the `latest` tag.

ctnkaan:
I've made some improvements but could not quite test it well

.github/workflows/release.yml
```diff
@@ -26,8 +26,19 @@ jobs:
...
+          echo "::set-output name=version::$VERSION"
+
       - name: Publish to NPM
         run: npm publish
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
           NPM_CONFIG_PROVENANCE: true
+
+      - name: Create GitHub Release
+        run: gh release create ${{ steps.get_version.outputs.version }} --title ${{ steps.get_version.outputs.version }} --notes "Full changelog at https://github.com/JedWatson/classnames/blob/main/CHANGELOG.md"
```
jonkoops, line 42: Here there is a reference to a step with the id `get_version`, but looks like the package version step is missing that id.

jonkoops, line 42: Thinking that we might want to use the current commit rather than `main`, so that the relevant notes for that release are always on top. WDYT?

ctnkaan: Oh you're right I'll fix it

ctnkaan: Sounds good I'll try to add this feature. By the way sorry for the late responses the work has been pretty heavy these few weeks.

jonkoops: No worries, I also maintain this stuff in my spare time, so I know the pain :wink:

0xD3n: I don't think it is a good idea to use the current commit as if the CHANGELOG did not get updated with that commit it would be empty so I thought linking the main branches CHANGELOG would be a better idea.

jonkoops: > I don't think it is a good idea to use the current commit as if the CHANGELOG did not get updated with that commit 

That's ok. I always update the changelog before release. In the future I'd like to automate this proccess also.

.github/workflows/release.yml
```diff
@@ -26,8 +26,19 @@ jobs:
           cache: npm
           registry-url: https://registry.npmjs.org
 
+      - name: Get package version
+        run: |
+          VERSION=$(npm pkg get version | xargs)
+          echo "Package version is $VERSION"
+          echo "::set-output name=version::$VERSION"
```
AlecRust, line 33:
```suggestion
          echo "name=$VERSION" >> $GITHUB_OUTPUT
```
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

jonkoops: Good catch.

OKEAMAH (approved):
Good

PR #379, opened 2024-02-28
feat: support Map and Set

ssi02014:
Thank you for providing such a great library. 🤗

I have a minor comment.
Are you considering support for `Set` and `Map`?
If so, I'd appreciate it if you could review whether this would be appropriate

I'm hoping that this will allow users to use the library in more ways.

But the opinion of maintainer is more important. 
If you think this is excessive or inappropriate, please let us know. Thank you!!! 😃

### DIFF
--- a/bind.js
+++ b/bind.js
@@ -26,18 +26,37 @@ function parseValue (arg) {
 		return classNames.apply(this, arg);
 	}
 
+	if (arg instanceof Set) {
+		return classNames.apply(this, Array.from(arg));
+	}
+
 	if (arg.toString !== Object.prototype.toString && !arg.toString.toString().includes('[native code]')) {
 		return arg.toString();
 	}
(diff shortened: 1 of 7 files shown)


jonkoops:
First off, thanks for the PR and starting a conversation.

We're currently in the proccess of simplifying our API, we want to reduce the amount of permutations that are accepted as input, so I don't feel this is something we'd want to do right now.

We could expand to accept every possible kind of data structure, but this utility has a very narrow goal of facilitating joining some classes conditionally, and I personally want to keep things simple. I also have no idea how this would affect performance, which is critical to our goals.

I'd also ask you, why would we need to accept `Set` and `Map`? What are the advantages? What practical use-cases would exist for this in the wild?

Closing for now, as I don't see us landing this at this time.

PR #303, opened 2023-03-24
Update README examples to use functional components

JoeDGit:
Suggestion: Updates examples to functional components

PR #388, opened 2024-04-11
Update README code space and variable declarations

wozitto:
## 1. Changed code space from tab to two spaces
The space in the code block has been replaced by a tab from two spaces in the PR [here](https://github.com/JedWatson/classnames/pull/303).
When I look at the space with tabs in preview, I feel the space is too wide, so I will change it back to two spaces.

- Current preview
スクリーンショット 2024-04-11 11 38 07

- Preview after changes
スクリーンショット 2024-04-11 11 41 23

## 2. Changed `let` declarations to `const` declarations in the "Dynamic class names with ES2015" example
Since the `const` declaration has been available from ES2015, I have used it to maintain consistency with other implementation examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,24 +83,24 @@ One of its primary use cases is to make dynamic and conditional `className` prop
 import React, { useState } from 'react';
 
 export default function Button (props) {
-	const [isPressed, setIsPressed] = useState(false);
-	const [isHovered, setIsHovered] = useState(false);
+  const [isPressed, setIsPressed] = useState(false);
+  const [isHovered, setIsHovered] = useState(false);
 
-	let btnClass = 'btn';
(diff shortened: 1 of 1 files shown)


dcousens:
This project seems to using tabs typically, but I don't really mind when 2 space is relatively standard in Javascript and this is only documentation

wozitto:
@dcousens 
Indeed, whether it is Tab or 2 space is a trivial difference.
However, since "Real-world example:" uses 2 space, I think it would be better to unify them for the sake of document consistency.

jonkoops:
I think this is a good thing to land, first and foremost because our documentation already uses inconsistent indentation in code examples. Secondarily because be default GitHub indentation for tabs is quite wide.

jonkoops (approved):

PR #392, opened 2024-04-17
fix: new @types/react support

Netail:
Support new @types/react version (v18.2.78 and higher)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -7,10 +7,10 @@
 //   Sean Kelley <https://github.com/seansfkelley>
 //   Michal Adamczyk <https://github.com/mradamczyk>
 //   Marvin Hagemeister <https://github.com/marvinhagemeister>
-export type Value = string | boolean | undefined | null;
+export type Value = string | boolean | undefined | null | number | bigint;
 export type Mapping = Record<string, any>;
-export interface ArgumentArray extends Array<Argument> {}
-export interface ReadonlyArgumentArray extends ReadonlyArray<Argument> {}
+export type ArgumentArray = Argument[];
+export type ReadonlyArgumentArray = ReadonlyArray<Argument>;
 export type Argument = Value | Mapping | ArgumentArray | ReadonlyArgumentArray;
 /**
  * A simple JavaScript utility for conditionally joining classNames together.


jonkoops:
Could you elaborate why this is needed? If you don't provide a reproducible example of the bug then it's hard to judge if these changes are needed.

Netail:
> Could you elaborate why this is needed? If you don't provide a reproducible example of the bug then it's hard to judge if these changes are needed.

@types/react added bigint support, which currently seems to break typescript when building;
```
 Type error: Argument of type '[string, string | false | 0 | 0n | null | undefined, string | undefined, string | false | undefined, string | undefined]' is not assignable to parameter of type 'ArgumentArray'.
   Type 'string | false | 0 | 0n | null | undefined' is not assignable to type 'Argument'.
     Type '0n' is not assignable to type 'Argument'.
 
   75 | }) => {
   76 |     const classes = classnames(
 > 77 |         css.root,
      |         ^
```
As classnames does not support this type

jonkoops:
I don't think we ever explicitly supported passing in a `ReactNode`, what would be the expectation of that? Could you provide a minimal example of passing a `ReactNode` in a practical use-case?

Netail:
> I don't think we ever explicitly supported passing in a `ReactNode`, what would be the expectation of that? Could you provide a minimal example of passing a `ReactNode` in a practical use-case?

Let's say you want change some css based on if the children are passed onto the component;

```
import type { FC, PropsWithChildren } from 'react';
import classNames from 'classnames';
import css from './test.module.scss';

export const Test: FC<PropsWithChildren> = ({ children, className }) => {
    const classes = classNames(
        css.root,
        children && css.hasChildren,
        className,
    );

    return (
        <div className={classes}>
            <h1>Hello</h1>
            {children}
        </div>
    )
}
```

jonkoops:
Thanks, I see your point. Since we have supported this historically I think we should land this, however this should be redirected to the `v2` branch, as we intend a different API for `v3` in the future, and it will likely not be released for a while.

Netail:
> Thanks, I see your point. Since we have supported this historically I think we should land this, however this should be redirected to the `v2` branch, as we intend a different API for `v3` in the future, and it will likely not be released for a while.

We can merge it into main and then cherry pick this change into the v2 branch

jonkoops:
No, this has to go into the v2 branch. The typings are correct for v3, and permutations there are likely to be reduced even further.

Netail:
> No, this has to go into the v2 branch. The typings are correct for v3, and permutations there are likely to be reduced even further.

Will create a new PR for that :) As changing the target branch pulls the commits in

Issue #393, opened 2024-04-19
Unable to do v2 releases


jonkoops:
@JedWatson could you please enable releases from the `v2` branch? See #364

dcousens:
@jonkoops what if we release from `main` anyway?

jonkoops:
Yeah, we could do that. We'd have to push what we have on `main` now to `v3` branch, and then push the `v2` branch to `main`. We'll still be in trouble if we want to continue to maintain `v2` in the future, but hopefully @JedWatson has given us the keys to the castle by then.

The only problem might be that there could be some branch protection set up for `main` that would prevent this. WDYT?

BearCooder:
So this seems to be basically not maintainable because you dont have the proper access? @jonkoops @JedWatson

jonkoops:
It blocks us from doing a v3 release and maintain a v2 at the same time, yes.

BearCooder:
He is pretty active on his Twitter that is linked in his github. maybe a DM on Twitter would help. I do not have an account there though

PR #394, opened 2024-04-20
fix: new @types/react support

Netail:
target v2 branch; https://github.com/JedWatson/classnames/pull/392

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -9,7 +9,7 @@
 //   Marvin Hagemeister <https://github.com/marvinhagemeister>
 
 declare namespace classNames {
-  type Value = string | number | boolean | undefined | null;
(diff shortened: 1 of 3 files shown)


jonkoops (changes requested):
Could you also add some test cases with bigint to the type tests and unit tests?

jonkoops (changes requested):
LGTM, but let's get rid of the unrelated formatting changes.

jonkoops (approved):
Looks great thanks @Netail!

jonkoops:
Just FYI, v2 releases are currently blocked on #393. I might switch `v2` back to `main` and `main` to `v3`, so we can perhaps push out a release. @dcousens WDYT?

PR #447, opened 2024-12-06
Add side effects to package.json

denisx:
Add side effects to package.json, to help bundlers cut unused code. Good with esm, modules, named exports...

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "url": "git+https://github.com/JedWatson/classnames.git"
   },
   "type": "module",
+  "sideEffects": false,
   "main": "./index.js",
   "types": "./index.d.ts",
   "exports": {


jonkoops:
Is there any evidence that this actually improves tree shaking for this package? We already provide ESM based entrypoints in the `exports` field, so I am pretty sure this will work in most bundlers without trouble.

denisx:
ESM is how files and imports are organized, Side effects is how a bundler can remove obviously unused data from an assembly. Only both make maximum effect.

rollup https://rollupjs.org/configuration-options/#treeshake-modulesideeffects
webpack https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free

jonkoops:
I understand the difference between imports and side effects, what I mean is that I don't think we need this metadata in order for bundlers to eliminate dead code. The WebPack documentation you linked even goes into this:

> In a 100% ESM module world, identifying side effects is straightforward. However, we aren't there quite yet, so in the mean time it's necessary to provide hints to webpack's compiler on the "pureness" of your code.

This package is now authored in ESM on the `main` branch, so I don't believe it will be required to include this.

denisx:
if your default import contain 100% needed code (and another files-imports this way), yes, this is not necessary settings.
but during refactoring yo can make effect situation...
ok, as you like. good day!

jonkoops:
Thanks for the suggestion, but I don't see a need to add this right now.

PR #453, opened 2024-12-31
Add type: bigint

ynifamily3:
I found that `bigint` type was added to @types/react.

It would be nice to support bigint for classNames types as well.

Usage:
```
  const reactNode: ReactNode = Math.random() ? 0n : <span>hello</span>;
  classNames(reactNode && 'foo');
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -7,7 +7,7 @@
 //   Sean Kelley <https://github.com/seansfkelley>
 //   Michal Adamczyk <https://github.com/mradamczyk>
 //   Marvin Hagemeister <https://github.com/marvinhagemeister>
-export type Value = string | boolean | undefined | bigint | null;
+export type Value = string | boolean | number | bigint | undefined | null;
 export type Mapping = Record<string, any>;
 export interface ArgumentArray extends Array<Argument> {}
 export interface ReadonlyArgumentArray extends ReadonlyArray<Argument> {}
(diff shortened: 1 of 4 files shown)


index.d.ts
```diff
@@ -7,7 +7,7 @@
 //   Sean Kelley <https://github.com/seansfkelley>
 //   Michal Adamczyk <https://github.com/mradamczyk>
 //   Marvin Hagemeister <https://github.com/marvinhagemeister>
-export type Value = string | boolean | undefined | null;
```
Phryxia, line 10: This is not related to mapped types in your modification of test code.

```js
classnames({
  a: 0,
  b: null,
})
```

The type of above is resolved as `Mapping`, not `Value`. For `Value` case, there is already nice test for it in [here](https://github.com/JedWatson/classnames/pull/453/files#diff-9c265283de2ab902547ac9c10806cee40c019f48d842ac7e8e014f868ca8cf9cL19).

I think `classnames` is meant to be used as CSS selector utilities, and class selector must be start with [#identifier](https://www.w3.org/TR/css-syntax-3/#ident-token-diagram) followed by `.`. Therefore it's nonsense to use numeric value as class selector, though using escape may allow it [#example](https://stackoverflow.com/questions/45293534/css-class-starting-with-number-is-not-getting-applied)

PR #565, opened 2026-06-19
docs: add Deno to install instructions

bartlomieju:
👋 Bartek from the Deno team here.

The README shows `npm install classnames`. Since Deno is a drop-in replacement for npm these days, this adds a Deno line alongside it:

```sh
deno install classnames
```

Docs-only change (classnames resolves and installs the same way under Deno). Totally fine to close this if you'd rather keep it npm-only. Happy to adjust.

_Disclosure: I work on Deno, so I have an interest here. The goal is parity with the package managers you already list, not promotion. This PR was prepared with AI assistance under human supervision: I review every change myself and personally respond to any comments or review feedback._

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,8 @@
 Install from the [npm registry](https://www.npmjs.com/) with your package manager:
 ```bash
 npm install classnames
+# or
+deno install classnames
 ```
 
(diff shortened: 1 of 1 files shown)


jonkoops:
I'd rather not add alternative package manager instructions, as this opens up the floodgate for all the others (and there are quite a lot). Thanks for the PR though.

PR #569, opened 2026-07-22
docs: document conditional string classNames pattern

### LINKED ISSUE #511: Very useful but not documented
`classNames('foo', true && 'bar'); // => 'foo bar'`

Hashim1999164:
## Summary
- Documents the common `condition && 'class'` pattern in the Usage examples.
- Fixes #511

## Test plan
- [ ] Confirm the new README examples match existing falsy-value behavior
- [ ] Skim Usage section for consistent style with nearby examples


jonkoops (approved):
Agreed, thanks!

